PR #4, opened 2015-06-14
Fix README

derekchiang:
Since it was copied directly from Google docs there are some formatting errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,38 +12,38 @@ This proposal introduces the concept of a command-line build tool, `acbuild`, th
 
 `acbuild` will support several commands:
 
-* acbuild init in.aci
+* `acbuild init in.aci`
   creates an empty aci image `in.aci` with manifest filled up with auto generated stub contents
 
-* acbuild env -var="a=b” -var=”c=d" -in=in.aci -out=out.aci   
+* `acbuild env -var="a=b" -var="c=d" -in=in.aci -out=out.aci`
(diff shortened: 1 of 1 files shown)


derekchiang:
Added a fix for 1) a typo (`in.aci` instead of `int.aci`) and 2) a wrong command (`add-image` instead of `add`)

README.md
```diff
@@ -164,8 +164,8 @@ build mongodb from official images
 use apt-get to install nginx
 
 
-    acbuild -in=in.aci -out=int.aci add aptitude.aci 
-    acbuild -in=int.aci -out=out.aci exec apt-get -y install nginx
+    acbuild -in=in.aci -out=int.aci add-image aptitude.aci
+    acbuild -in=in.aci -out=out.aci exec apt-get -y install nginx
```
ecnahc515: the `-in` image should be `int.aci`, the output of the previous call, yes?

klizhentas: right, int.aci is just a typo

ecnahc515: Then the `-out=int.aci` in the first acibuild command should be `-out=in.aci`?

jonboulle:
@derekchiang thanks for the patch! would you mind squashing your commits?

derekchiang:
@jonboulle done

README.md
jonboulle: actually, int was not a typo, it was intended to be an intermediate ACI.

Issue #5, opened 2015-06-19
*: initial implementation


jonboulle:
The [README](https://github.com/appc/acbuild/blob/master/README.md) describes what acbuild should look like; now we need an initial implementation of the thing.

sgotti:
@jonboulle Is someone (@klizhentas ?) already working on it?

klizhentas:
we chatted with @derekchiang about hacking on this and outlined a simple plan, but haven't started yet. @sgotti you want to hack together?

sgotti:
@klizhentas @derekchiang Good to know! Perhaps, due to other work in this week (I don't want to become a bottleneck) it'll be better if we can just put a basic initial implementation PR of some initial commands and then follow with the classic workflow opening various PR.

I'm open for helping and for every suggestion!

klizhentas:
@sgotti agreed. We will come up with the initial implementation PR for the most important command - exec in the next couple of weeks, and we'll go from there.

derekchiang:
Sounds good to me. I have been busy this week but I will start working on
it over this weekend.

dgonyeo:
An initial implementation was just merged.

PR #10, opened 2015-08-21
Initial implementation

derekchiang:
Tracking #5.


sgotti:
@derekchiang thanks! I just started an high level review.
I'll skip talking about the frameworks used (like logrus instead of the standard logger or runc for exec) as they can be discussed in future.

If I'm not wrong by default images aren't squashed. I'm ok with this as it will speedup working with an image that depends on a big image since it doesn't need to write big acis for every command. But acbuild should correctly write image dependencies (using also the labels) or, when executed by an ACE (like rkt) it will need to discover and fetch them but probably won't have enough information to find the right image to compose the discovery url or it may download the wrong image (since imageID is provided in the image dependency info, the ACE will fail if the wrong image is discovered).

This also means that `render` and `squash` (now I don't see it implemented) will work only if the dependencies are already in the acbuild store or it will need to download them (but this can be addressed in the future).

jonboulle:
@derekchiang do you have any time to come back to this? I'd love to pick it up and get it moving again. Thanks!

klizhentas:
@jonboulle  Last time I checked acbuild was working - so we should kick off a good code reviewing session first to send the feedback to @derekchiang

klizhentas:
oh, I missed a good code review from @sgotti here too

derekchiang:
@sgotti thank you so much for the code review, and sorry for the late reply!  I'm quite busy right now but I do intend to go over the code review, probably over this weekend.  Once the code is in a better shape, @jonboulle could probably take over, as he suggested.

derekchiang:
@sgotti I've replied to your comments and partially addressed them in code.  I think the next big thing is to port the image discovery/fetching logic from `rkt` into `acbuild`.  Should we merge this and use other PRs for future work?

sgotti:
@derekchiang thanks! I answered to your question and added few other comments. Other than my doubts on `acb add` options I think that future improvements (like extracting common labels like version/os/arch from dependencies to help an ACE discovering the right image and consequently rkt fetch logic, contexts etc...) can be done in future PRs.
Probably, the README should be updated with the new commands names (and perhaps a few notes on what is/isn't implemented)

I'd also like to hear other reviews/opinions (@jonboulle?) on this.

jonboulle:
I've asked @dgonyeo to take a look at this as I'm a little swamped at the moment but I do hope to make a pass in the next week..

dgonyeo:
```
sudo ./acb render foo.aci foo
ERRO[0000] error rendering image in store: error rendering tree store: treestore: cannot render aci: no keys found
```
Does this happen because foo.aci isn't signed?

dgonyeo:
Just as a tiny nitpick, the Usage section of the top level help (`acb --help`) seems weird:

```
Usage: 
  acb [flags]
  acb [command]
```

I'd vote to remove the `acb [flags]` line. I glanced at the code quickly, but I'm not sure where in there it's coming from.

dgonyeo:
`acb deps` prints the following error when run as root (but works when run as non-root):

```
sudo ./acb deps foo2.aci
ERRO[0000] error creating dependency tree: error processing dependencies for image (foo2): error resolving image (centos) to key: aci not found
```

`foo2.aci` was built via the following steps:

```
sudo rkt fetch --insecure-skip-verify docker://centos
sudo rkt image export sha512-1edd13e7ce85a6f4f81e83397b81f5d1d897ac8bc7613eca00dd36a5925a248b centos.aci
acb new foo.aci -n foo
acb add centos.aci foo.aci -o foo2.aci -n foo2
```

dgonyeo:
acb lets me make circular dependencies. Perhaps it should try to detect this, and not allow this?

```
acb new foo.aci -n foo
acb add foo.aci -o foo.aci -n foo
acb deps foo.aci
```

If the above commands are run the last one will hang.

dgonyeo:
Related to my last comment, perhaps `acb deps` should detect when there is a circular dependency, and print an error?

sgotti:
> Does this happen because foo.aci isn't signed?

It's probably happening because when running as root it's using a diferent store placed in the root's home instead of the one in the user home.

Since there're some commands (`render` needs root for chroot caps and for creating file with different users/groups, and probably `exec`/`runc` needs all the caps for spawning a container) that needs to be run as root and other that can be executed as a normal user this is quite confusing. 

As acbuild is using the rkt store probably acbuild will need to follow the same path (coreos/rkt#1452).

dgonyeo:
Closing this, as the other implementation was merged.

PR #11, opened 2015-09-29
Added initial implementation

dgonyeo:
There's still a good amount of work to be done on parts of this, but I think it's reached a usable state so I suppose it's a good time to kick this up for review.

I've deviated some from the README that was written. Notable differences:
- Context-free mode isn't implemented, and the contextual mode behaves a little different
- Exec only supports an un-layered build with overlayfs (overlayfs not required if there are no dependencies)

With this implementation the intermediate ACI is stored uncompressed and untarred until `acbuild finish` is called. This results in good performance for every command except `init` and `finish` regardless of ACI size.

At this stage the thing I'm least happy about is the way it stores dependencies, and will focus on improving that next. The current method feels very brittle. There are also parts around dependency fetching that need work.

An example build script with > 5 steps is available [here](https://github.com/dgonyeo/containers/blob/master/rkt/plex/build-plex.sh).


dgonyeo:
@sgotti your opinions on this would be appreciated

sgotti:
@dgonyeo Great! I'll take a look at it tomorrow.

BTW 
* I think it'll need to be run as root for some commands (to render and exec) right?

* Have you thought about using the acirenderer from appc/spec to render the ACIs (it handles the deps, pathwhitelist and it's already used by rkt and docker2aci to render it without the need to extract anything to filesystem)?

dgonyeo:
I glanced at it, but I didn't quite understand how I would use it. I'll talk to @jonboulle about how I should incorporate it.

And yes, root is required for exec.

acbuild/add-anno.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/appc/acbuild/libacb"
+)
+
+var (
+	cmdAddAnno = &cobra.Command{
+		Use:     "add-anno NAME VALUE",
```
ecnahc515: Why not `acbuild annotation add`?

acbuild/finish.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/appc/acbuild/libacb"
+)
+
+var (
+	actool string
+
+	flagFinOverwrite = false
+	cmdFinish        = &cobra.Command{
+		Use:     "finish ACI_PATH",
```
ecnahc515: maybe a different word than 'finish'? Seems like a bad choice of verbiage.

libacb/util/deps.go
```diff
@@ -0,0 +1,354 @@
...
+
+var (
+	defaultArch = "amd64"
+	defaultOS   = "linux"
+)
+
+// RenderACI will download any missing dependencies into the depstore,
+// expand any unexpanded dependencies into the scratchpath, and return
+// an ordered list of all dependencies for overlaying.
+func RenderACI(acipath, scratchpath, depstore string, insecure bool) ([]string, error) {
```
sgotti: This is because I was suggesting to use appc/spec acirenderer. Overlaying dependencies is not enough since aci may have a pathWhiteList. I'll be happy to do/help you moving this to acirenderer. Let me know if you prefer to do this in this PR or in a follow up PR.

libacb/util/deps.go
```diff
@@ -0,0 +1,354 @@
...
+	return cmd.Run()
+}
+
+//TODO: path should include labels probably
+func GenDepPath(depstore, imageName string) string {
+	return path.Join(depstore, strings.Replace(imageName, "/", "-", -1)+".aci")
+}
+
+func FetchDeps(deps types.Dependencies, depstore string, insecure bool) error {
+	for _, dep := range deps {
```
sgotti: In future I'll move this the rkt fetching logic and avoid duplicating some work bugfixes (it should be moved, as already talked somewhere, to its own package)

libacb/util/deps.go
```diff
@@ -0,0 +1,354 @@
...
+func VerifyACIHash(path string, dephash types.Hash) (bool, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return false, err
+	}
+
+	var h hash.Hash
+	typ := strings.SplitN(dephash.String(), "-", 2)[0] // this is silly
+	switch typ {
+	case "sha256":
```
sgotti: Actually app/spec only support `sha512.

libacb/finish.go
```diff
@@ -0,0 +1,79 @@
...
+			return fmt.Errorf("ACI already exists: %s", output)
+		}
+		err := os.Remove(output)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = util.Exec(actool, "build", "--overwrite", tmpaci, output)
+	if err != nil {
```
sgotti: what about using appc/spec/aci ImageWriter instead of calling an external process?

acbuild/acbuild.go
```diff
@@ -0,0 +1,93 @@
...
+	cliDescription = "acbuild, the application container build system"
+)
+
+var (
+	tmpaci       = ".acbuild.tmp"
+	targetpath   = ".target"
+	scratchpath  = ".scratch"
+	depstorepath = ".depstore"
+	workpath     = ".work"
+
```
sgotti: These are created in the cwd right? So you have context mode for free and can have multiple build just executing acbuild in different dirs right? I like it.

* What about using a parent dir (something like `.acbuild`) to contain them?
* What about an option to specify the path of the build (if not specified will default to the cwd)

acbuild/add-anno.go
sgotti: About `command object` vs `object command` I don't have a strong opinion. Given that the command that works on objects are just `add` and `rm` while the objects can be a a lot it'll be more easy to group the objects under them...
With cobra it's also easy to add subcommands.

acbuild/finish.go
dgonyeo: I've renamed `init` to `begin` and `finish` to `end`.

libacb/finish.go
dgonyeo: That makes a lot more sense! Just pushed a commit that switches to that.

acbuild/acbuild.go
dgonyeo: I like both your suggestions, just finished implementing them.

libacb/util/deps.go
dgonyeo: I'll switch to acirenderer, and I see no reason to not tack it onto this PR.

klizhentas:
looks great - as a side note, Eventually It would be still nice to get a context free mode for environment variables patching, and any other simple one-two command modifications, may be in the next iterations

dgonyeo:
I do intend to add a context free mode, I don't imagine it'll be too hard to implement.

dgonyeo:
At @jonboulle's prompting, here's a short summary of some of the significant differences between this implementation and the one that was proposed earlier:
- Uses systemd-nspawn instead of runc
- Doesn't use rkt's store (or have any dependency on rkt)
- Dependency store is per-context, and specific to the current working directory
- Strong focus on having a build context, a context-free mode will probably just be a wrapper around making a context, running a single command, and removing the context
- Won't exec if there are dependencies and overlayfs isn't supported (no way to pull out changes to the ACI being built without squashing it with all of its dependencies)

jonboulle:
> - Uses systemd-nspawn instead of runc
> - Doesn't use rkt's store (or have any dependency on rkt

I am much more inclined to go with this direction for these two reasons alone. I'm also a fan of the context-by-default approach.

@dgonyeo would you say this is ready for review or still some further changes you plan on making first?

dgonyeo:
I want to switch to using acirenderer from the appc/spec repo as per sgotti's suggestion, so anything related to ACI rendering isn't ready for review (mostly in `libacb/util/deps.go`), but the rest of the codebase is ready for nitpicking.

klizhentas:
@dgonyeo @jonboulle I'd be interested in having an option to use `runc` as a pluggable engine - makes it easier not to drag dependency on `systemd-nspawn` in this tool. We can add the feature if you are ok with it.

jonboulle:
Yeah I guess that's reasonable.

dgonyeo:
You can also say `anno add` instead of `annotation add` if you want to save yourself the typing. The help pages may not highlight this well enough though.

PR #12, opened 2015-10-05
Added initial implementation

dgonyeo:
Jon asked me to rebase my changes, when I did github closed the last PR (https://github.com/appc/acbuild/pull/11) and it won't let me reopen it, so here's a new PR with the rebased changes.

The main change since the last round of reviews is that [acirenderer](https://github.com/appc/spec/tree/master/pkg/acirenderer) is now being used to render the ACIs.


jonboulle:
A couple of high-level comments:
- how do you intend to handle concurrent invocations of `acbuild`? (it almost certainly makes zero  sense as a use case, but we need to at least block it from breaking things)
- re: package naming/organisation - we don't have a great track record of consistency in this area across CoreOS projects (I think because we've never really settled on something we're quite happy with) in terms of where we put commands, libs, etc. I would probably rather just `s/libacb/lib` for now - the import path means it's still unambiguous and it's what we do in docker2aci.

dgonyeo:
Re concurrent invocations:

Concurrent invocations in different working directories will be fine, but I assume you mean concurrent invocations in the same working directory. To prevent this being an issue, `acbuild` could get a lock on a file in `.acbuild.tmp` before doing any operation, and release it when done.

Re package reorg:

Will do.

jonboulle:
> Concurrent invocations in different working directories will be fine, but I assume you mean concurrent invocations in the same working directory. To prevent this being an issue, acbuild could get a lock on a file in .acbuild.tmp before doing any operation, and release it when done.

Right, that is what I mean. We can address that in a follow up - I would like to get this landed

jonboulle:
Please move /libacb/* --> /*, no need for the extra level.

Other than that, LGTM, let's get this merged and we can iterate from there. Awesome work!

jonboulle:
boom!

ecnahc515:
Huzzah! :+1:

iaguis:
\o/

Issue #14, opened 2015-10-07
*: Use file locking to prevent concurrent invocations


dgonyeo:
Concurrent invocations in acbuild's current state can cause issues. A file lock should be acquired/released in the build context's directory to prevent this.

PR #26, opened 2015-10-07
*: Implemented file locking for each command

dgonyeo:
Each command now acquires a file lock on a file in the current working
context before performing any actions, and releases it upon completion.
This prevents concurrent invocations of acbuild from producing
unintended side effects.

### DIFF
--- a/acbuild/abort.go
+++ b/acbuild/abort.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"os"
 	"path"
 
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
@@ -42,14 +43,29 @@ func runAbort(cmd *cobra.Command, args []string) (exit int) {
 		return 1
 	}
 
+	lockfile, err := getLock()
+	if err != nil {
+		stderr("abort: %v", err)
+		return 1
+	}
+	// Lock will be released when lib.Abort deletes the folder containing the
+	// lockfile.
+
 	if debug {
 		stderr("Aborting the build")
 	}
 
-	err := lib.Abort(path.Join(contextpath, workprefix))
+	err = lib.Abort(path.Join(contextpath, workprefix))
 
 	if err != nil {
 		stderr("abort: %v", err)
+		// In the event of an error the lockfile may have not been removed, so
+		// let's release the lock now
+		if err := releaseLock(lockfile); err != nil {
+			if !os.IsNotExist(err) {
+				stderr("abort: %v", err)
+			}
+		}
 		return 1
 	}
 
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -19,11 +19,14 @@ import (
 	"os"
 	"path"
 	"strings"
+	"syscall"
 	"text/tabwriter"
 	"text/template"
 
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/pflag"
+
+	"github.com/appc/acbuild/util"
 )
 
 const (
@@ -72,6 +75,10 @@ func depstorepath() string {
 
 func workpath() string {
 	return path.Join(contextpath, workprefix, "work")
+}
+
+func lockpath() string {
+	return path.Join(contextpath, workprefix, "lock")
 }
 
 var cmdAcbuild = &cobra.Command{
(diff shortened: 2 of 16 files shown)


dgonyeo:
This should resolve https://github.com/appc/acbuild/issues/14

acbuild/acbuild.go
```diff
@@ -156,6 +163,41 @@ func stdout(format string, a ...interface{}) {
...
+	if err != nil {
+		return err
+	}
+
+	lockfile, err = os.OpenFile(lockpath(), os.O_CREATE|os.O_RDWR, 0644)
+	if err != nil {
+		return err
+	}
+
+	err = syscall.Flock(int(lockfile.Fd()), syscall.LOCK_EX)
```
jonboulle: You're going to want LOCK_NB here, and probably to propagate a nicer error in the case that it's already locked.

acbuild/annotation.go
```diff
@@ -83,16 +96,29 @@ func runRmAnno(cmd *cobra.Command, args []string) (exit int) {
...
 	}
 
-	err := lib.RemoveAnnotation(tmpacipath(), args[0])
+	err = lib.RemoveAnnotation(tmpacipath(), args[0])
 
+	var retcode int
 	if err != nil {
 		stderr("annotation remove: %v", err)
-		return 1
+		retcode = 1
```
jonboulle: Defer the release just after you get the lock, then you don't need to worry about these differing return paths

jonboulle: Then again we could really just leave this implicit. How about we revert to just letting the lock be removed when the process gets turned down

jonboulle: OR actually remove the lockfile in releaseLock....

acbuild/acbuild.go
dgonyeo: Why not just wait for the lock to be released?

jonboulle: I dunno, I'd consider any situation where there are concurrent invocations
to be a bug (in whatever is invoking acbuild) and for it to fail fast.
Otherwise it seems inherently racy; for the most part acbuild operations
are necessarily sequential. Can you think of a use case I'm missing?

acbuild/annotation.go
dgonyeo: I'd prefer to clean up after myself. Not removing the lockfile is pretty irrelevant to acbuild, but from a UX standpoint I suppose the user would expect the file to not exist when acbuild doesn't have a lock on it. I'd also prefer to not defer the release, since if the lock can't be released for whatever reason I'd really like to emit an error.

I suppose there's a chance someone (possibly myself when writing tests) would directly call the cmd* functions, and thus couldn't rely on the lock being released when the process exits.

jonboulle: > Not removing the lockfile is pretty irrelevant to acbuild, but from a UX standpoint I suppose the user would expect the file to not exist when acbuild doesn't have a lock on it. 

Right I think it would be cleaner to remove the file in the usual case

> I'd also prefer to not defer the release, since if the lock can't be released for whatever reason I'd really like to emit an error.

I meant deferring a closure/inline func, not just the releaseLock() call. You can still emit and propagate errors.
 
> I suppose there's a chance someone (possibly myself when writing tests) would directly call the cmd* functions, and thus couldn't rely on the lock being released when the process exits.

I think it's perfectly reasonable to not support running the cmd functions directly from code, they should be the CLI glue and may do exits/direct stdout/stderr IO/blabla

acbuild/acbuild.go
dgonyeo: I suppose not. I'll make the change.

acbuild/abort.go
```diff
@@ -42,11 +42,23 @@ func runAbort(cmd *cobra.Command, args []string) (exit int) {
 		return 1
 	}
 
+	err := getLock()
```
jonboulle: could you actually just have this return the lock, and then pass that into the defer func. I'm not a big fan of the global variable. Sorry I didn't raise this before..

dgonyeo: Yup

dgonyeo:
This should be ready for another pass when you get the time.

acbuild/set-exec.go
```diff
@@ -40,11 +40,23 @@ func runSetExec(cmd *cobra.Command, args []string) (exit int) {
...
 	}
 
+	lockfile, err := getLock()
+	if err != nil {
+		stderr("set-exec: %v", err)
+		return 1
+	}
+	defer func() {
+		if err := releaseLock(lockfile); err != nil {
+			stderr("run: %v", err)
```
jonboulle: set-exec

acbuild/copy.go
```diff
@@ -43,11 +43,23 @@ func runCopy(cmd *cobra.Command, args []string) (exit int) {
...
 	}
 
+	lockfile, err := getLock()
+	if err != nil {
+		stderr("copy: %v", err)
+		return 1
+	}
+	defer func() {
+		if err := releaseLock(lockfile); err != nil {
+			stderr("begin: %v", err)
```
jonboulle: copy

jonboulle:
A couple of small copypastas, but lgtm after that

acbuild/acbuild.go
```diff
@@ -156,6 +163,52 @@ func stdout(format string, a ...interface{}) {
...
+	}
+
+	lockfile, err := os.OpenFile(lockpath(), os.O_CREATE|os.O_RDWR, 0644)
+	if err != nil {
+		return nil, err
+	}
+
+	err = syscall.Flock(int(lockfile.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)
+	if err != nil {
+		if err.Error() == "resource temporarily unavailable" {
```
jonboulle: if errno, ok := err.(syscall.Errno); ok && errno == syscall.EWOULDBLOCK

jonboulle: hmm or...
https://github.com/coreos/rkt/blob/master/pkg/lock/dir.go#L50

dgonyeo: that's way better

acbuild/abort.go
```diff
@@ -42,14 +42,27 @@ func runAbort(cmd *cobra.Command, args []string) (exit int) {
...
 	}
 
-	err := lib.Abort(path.Join(contextpath, workprefix))
+	err = lib.Abort(path.Join(contextpath, workprefix))
 
 	if err != nil {
 		stderr("abort: %v", err)
+		// In the event of an error the lockfile won't have been removed, so
+		// let's release the lock now
+		if err := releaseLock(lockfile); err != nil {
```
jonboulle: This is a bit pedantic, but: it's possible for Abort() to fail, yet for it to have successfully removed the lockfile (for example, if the RemoveAll fails partway through after it has removed the lockfile), in which case the releaseLock will fail somewhat spuriously. One possible solution would be for releaseLock to not fail if its os.Remove() returns os.ErrNotExist.

acbuild/acbuild.go
```diff
@@ -156,6 +163,52 @@ func stdout(format string, a ...interface{}) {
...
+	lockfile, err := os.OpenFile(lockpath(), os.O_CREATE|os.O_RDWR, 0644)
+	if err != nil {
+		return nil, err
+	}
+
+	err = syscall.Flock(int(lockfile.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)
+	if err != nil {
+		if err == syscall.EWOULDBLOCK {
+			return nil, fmt.Errorf("lock already held - is another acbuild running in this working dir?")
+		} else {
```
jonboulle: Nit, it's preferred style to just do short-circuit return - so, remove the "else"

jonboulle:
two more nits, sorry. then land it.

Issue #15, opened 2015-10-07
end: image signing


dgonyeo:
It would be really convenient to be able to end a build and have acbuild produce the `.asc` file in addition to the `.aci` file, if it's handed the keyrings. 

Something like the following:

```
acbuild end --keyring ./keys.pub --secret-keyring ./keys.sec mycoolapp-latest-linux-amd64.aci
```

Would produce both a `mycoolapp-latest-linux-amd64.aci` file and a `mycoolapp-latest-linux-amd64.aci.asc` file in the current directory.

PR #27, opened 2015-10-08
end: added flag for signing images

dgonyeo:
This adds support for having acbuild use gpg for you to sign the
produced aci. By default gpg is called with the following flags:

`--armor --yes --output mycoolapp.aci.asc --detach-sig mycoolapp.aci`

If any flags are provided to `end` after the path to the ACI, the
`--armor` and `--yes` flags are removed, and these provided flags are
prefixed to the list of flags passed to gpg.


dgonyeo:
Solves https://github.com/appc/acbuild/issues/15

dgonyeo:
I tried to use the openpgp golang library instead of exec'ing gpg, but I ran into too much frustration and then gave up.

lib/end.go
```diff
@@ -29,14 +29,14 @@ import (
...
-func End(tmpaci, output, contextpath string, overwrite bool) error {
+func End(tmpaci, output, contextpath string, overwrite, sign bool, gpgflags []string) error {
 	man, err := util.GetManifest(tmpaci)
 	if err != nil {
 		return err
 	}
 
 	if man.App != nil && testEq(man.App.Exec, placeholderexec) {
-		return fmt.Errorf("can't end build, run command was never set")
+		return fmt.Errorf("can't end build, exec command was never set")
```
jonboulle: Welp, this is no longer valid thanks to https://github.com/appc/spec/pull/505. Although I would still suggest we at leats print a warning in this case.

lib/end.go
```diff
@@ -63,13 +63,23 @@ func End(tmpaci, output, contextpath string, overwrite bool) error {
 	}
 
 	aw := aci.NewImageWriter(*man, tar.NewWriter(ofile))
-	defer aw.Close()
 
 	err = filepath.Walk(tmpaci, aci.BuildWalker(tmpaci, aw, func(hdr *tar.Header) bool { return true }))
```
jonboulle: won't passing nil to BuildWalker instead of the closure achieve the same thing?

acbuild/end.go
```diff
@@ -44,16 +46,13 @@ func runEnd(cmd *cobra.Command, args []string) (exit int) {
...
-		return 1
-	}
 
 	if debug {
 		stderr("Ending build. Writing completed ACI to %s", args[0])
 	}
 
-	err := lib.End(tmpacipath(), args[0], path.Join(contextpath, workprefix), overwrite)
+	err := lib.End(tmpacipath(), args[0], path.Join(contextpath, workprefix),
+		overwrite, sign, args[1:])
```
jonboulle: Your use of args[1:] is magical and undocumented, is that intentional? Do we need to expose this to users yet?

util/exec.go
```diff
@@ -23,6 +23,7 @@ import (
 // from the binary's stdout and stderr to this program's stdout and stderr.
 func Exec(bin string, args ...string) error {
 	cmd := exec.Command(bin, args...)
```
jonboulle: Consider using https://golang.org/pkg/os/exec/#LookPath as a nicety here (and returning a friendly error) instead of just going straight to the exec

dgonyeo: The `exec.Command` function does that for me: https://golang.org/src/os/exec/exec.go?s=3999:4044#L112

acbuild/end.go
dgonyeo: Yes it's intentional, and I want to expose it to users. I just need to document it better.

All arguments given to the end command after the path to the ACI to write are passed to gpg. This allows for the following:

```
acbuild end --sign thing3.aci -- --no-default-keyring --keyring test1.gpg
```

dgonyeo: I updated the example usage of `end` in the cobra help page, hopefully it'll be a tad clearer now. Maybe the REAME also deserves a paragraph explaining this.

dgonyeo: https://github.com/appc/acbuild/pull/27/files#diff-04c6e90faac2675aa89e2176d2eec7d8R175

jonboulle: Ehhh, I am wary of this approach because it precludes us from supplying other kinds of args. Can we have them as options passed to `--sign` rather than args[1:] ?

util/exec.go
jonboulle: So it does! I'm used to just raw Cmds. Carry on.

jonboulle: what does the error look like now?

lib/end.go
```diff
@@ -89,3 +99,14 @@ func testEq(a, b []string) bool {
 	}
 	return true
 }
+
+func signACI(acipath, signaturepath string, flags []string) error {
+	if flags == nil {
```
jonboulle:
```
if len(flags) == 0 {
    flags = append("--armor", "--yes", "--output", "--detach-sig", acipath)
}
return util.Exec("gpg", flags...)
```

jonboulle: len is preferable because of http://play.golang.org/p/M6w2YsK9B9

util/exec.go
dgonyeo: which error is this?

jonboulle: like, "gpg is required but not found - is it installed and on your PATH?" Something like that.

jonboulle: @dgonyeo general theme here is that acbuild should be as hand-holdy as possible, so prefer to wrap errors and provide suggestions

dgonyeo: How's [this](https://github.com/appc/acbuild/pull/27/files#diff-5601779b2e1aae16ccfdcd9bc9d17dc6R111)?

I also [updated run](https://github.com/appc/acbuild/pull/27/files#diff-e40b1e3cb0587929a3c786527046d69aR123) to have a similar check.

jonboulle: lgtm

jonboulle:
Lgtm, although thinking about it further we should probably fail as early as possible if any required execs aren't present; otherwise, what state is the .acbuild going to be in?

dgonyeo:
If the `gpg` command fails (where fails means had a non-0 exit code), an error will be returned, the ACI that was produced will be deleted (happens [here](https://github.com/appc/acbuild/pull/27/files#diff-5601779b2e1aae16ccfdcd9bc9d17dc6R80), just added it), and the current build context should be unchanged.

In other words, I think it'll be left in the state it was before `end` was called, and the build can then be continued, canceled, or end can be attempted again.

README.md
```diff
@@ -172,6 +172,25 @@ dependencies.
...
+
+The `--user` flag may also be specified for the user to run the `gpg` command
+as. This is provided to allow running `end` as root, which may be necessary to
+clean up files created by the root user during a `run` command, while still
+using the user's keyring.
+
+An example of this functionality is:
+
+```
+acbuild end --sign mynewapp.aci -- --no-default-keyring --keyring ./rkt.gpg
```
jonboulle: I think the example should have `--user` (or add another example) as I'd expect that to be more common

jonboulle: I'm almost wondering about splitting this out into two steps - `acbuild end` and `acbuild cleanup`. It just feels awkward to run an end command as root.

acbuild/end.go
```diff
@@ -61,7 +61,8 @@ func runEnd(cmd *cobra.Command, args []string) (exit int) {
 		stderr("Ending build. Writing completed ACI to %s", args[0])
 	}
 
-	err = lib.End(tmpacipath(), args[0], path.Join(contextpath, workprefix), overwrite)
+	err = lib.End(tmpacipath(), args[0], path.Join(contextpath, workprefix),
+		overwrite, sign, user, args[1:])
```
jonboulle: What if len(args) == 1?

lib/end.go
```diff
@@ -89,3 +101,31 @@ func testEq(a, b []string) bool {
...
+		if sucommand == "" {
+			sucommand = flag
+		} else {
+			sucommand = sucommand + " " + flag
+		}
+	}
+
+	var err error
+	if user == "" {
+		err = util.Exec("su", user, "-c", sucommand)
```
jonboulle: What I had in mind was actually using `cmd.SysProcAttr.Credential` on the gpg command rather than calling out to su. I am not sure if there is any other initialization that I'm missing that would prevent gpg working that way as another user; can you try it?

acbuild/end.go
dgonyeo: Then `args[1:]` will be an empty slice.

https://play.golang.org/

lib/end.go
dgonyeo: Sure

README.md
dgonyeo: I'll add an example for `--user`.

I think that would be fine, but I want to keep the final command called `end` since the first command is called `begin`. How about calling the command that produces an ACI `write`?

Splitting the two out would also make `end` do the exact same thing as `abort`, so that won't be needed anymore.

jonboulle: That sounds reasonable to me

dgonyeo:
Should be ready for another pass. `end` was renamed to `write`, and no longer removes the build context. `abort` was renamed to `end`, and is intended to be called at the end of a build, regardless of success or failure. This means that `write` doesn't need to be called as root (probably, unless there's directories that were created in the ACI it can't get into due to permissions).

An example bash script using the new semantics is [here](https://github.com/dgonyeo/containers/blob/master/rkt/blog/build-site.sh), if anyone wants to see what it would look like.

This also enables a user to save an ACI at multiple points in a build, if they want to do so for debugging purposes, which I think is neat.

README.md
```diff
@@ -72,13 +72,14 @@ The following commands are supported:
...
   Begin a build.
 
-* `acbuild end ACI_PATH`
+* `acbuild end`
 
-  End a build, writing the resulting ACI to `ACI_PATH`.
+  End a build, deleting the current build context.
 
-* `acbuild abort`
+* `acbuild write ACI_PATH`
```
jonboulle: Nit, I suggest listing `write` before `end` so it's more obvious that the normal workflow involves both.

acbuild/end.go
```diff
@@ -58,17 +52,19 @@ func runEnd(cmd *cobra.Command, args []string) (exit int) {
...
+	err = lib.End(path.Join(contextpath, workprefix))
 
 	if err != nil {
 		stderr("end: %v", err)
-		// In the event of an error the lockfile won't have been removed, so
+		// In the event of an error the lockfile may have not been removed, so
 		// let's release the lock now
 		if err := releaseLock(lockfile); err != nil {
-			stderr("end: %v", err)
+			if !os.IsNotExist(err) {
```
jonboulle: can just collapse this into line 64

acbuild/write.go
```diff
@@ -24,48 +23,52 @@ import (
...
 var (
-	cmdAbort = &cobra.Command{
-		Use:     "abort",
-		Short:   "Abort an existing build",
-		Long:    "Abort the current build, throwing away any changes since init was called",
-		Example: "acbuild abort",
-		Run:     runWrapper(runAbort),
+	overwrite = false
+	sign      = false
+	user      = ""
```
jonboulle: kill this now

jonboulle:
LGTM - @krobertson any chance of a review? :-)

krobertson:
LGTM. Two notes:

1) Readme example at the bottom isn't updated, it uses `acbuild end`.

2) Using `--overwrite`, it'll remove and recreate the aci, but not the signature. GPG doesn't error if the file already exists, so it will overwrite it fine. On a failed signing, acbuild will remove the aci, however it could leave an old signature behind.

jonboulle:
Good catches - @dgonyeo clean those up and merge?

jonboulle:
posthumous lgtm? :P

PR #29, opened 2015-10-10
copy: create parent directory before attempting to copy

dgonyeo:
For convenience, the copy command should create the parent directory for
the copy destination

### DIFF
--- a/lib/copy.go
+++ b/lib/copy.go
@@ -15,16 +15,26 @@
 package lib
 
 import (
+	"os"
 	"path"
 
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/aci"
-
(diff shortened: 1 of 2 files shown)


lib/copy.go
```diff
@@ -25,6 +26,15 @@ import (
 // Copy will copy the directory/file at from to the path to inside the untarred
 // ACI at acipath.
 func Copy(acipath, from, to string) error {
```
jonboulle: Can we move to this? https://github.com/coreos/rkt/blob/master/pkg/fileutil/fileutil.go#L63
I'm happy for the interface to change slightly if it makes it easier, I think the uid stuff might be a bit of a leak where a generic callback would be better.

jonboulle:
LGTM, but needs rebase before merge.

PR #30, opened 2015-10-12
README: fix command: "dependency add"

alban:
The command add-dep does not exist.


jonboulle:
I asked about this before and was under the impression there was some kind of cobra magic that actually made both work...

jonboulle:
Maybe I'm thinking of "dep add"

Issue #28, opened 2015-10-09
util: replace UnTar with an external library


dgonyeo:
There are issues with the UnTar in its current state, and it's not a particularly easy thing to write. It should be replaced with a library.

jonboulle:
Can you just use https://github.com/coreos/rkt/tree/master/pkg/tar ? if not, why not? Perhaps we could make it more generic and push it into github.com/coreos/pkg or so

jonboulle:
You're right that it's not an easy thing to write, and I'd really rather not have multiple implementations of it, but we're fairly happy with where the rkt one is right now.

dgonyeo:
I should be able to use that, but I had an issue or two when I tried to switch. I'll try again and get feedback from others if I have issues.

alban:
+1 for https://github.com/coreos/rkt/tree/master/pkg/tar: it should solve this problem: extracting character devices:
```
# acbuild run -- apt-get update
Downloading quay.io/fermayo/ubuntu: 72.4 MB of an unknown total size
run: unknown type 3 for file in tar: rootfs/dev/agpgart
```

jonboulle:
@dgonyeo doh! I realised the critical missing piece from your experiments the other day - we have to tell acbuild to try to multicall so that it can be invoked in the tar mode. Check out https://github.com/jonboulle/acbuild/commit/236d8965983a7b47c275a24ee11f966e81fd5957

jonboulle:
With that applied:
```
: trevize acbuild extracttar ; bin/acbuild --debug begin
Beginning build with an empty ACI
: trevize acbuild extracttar ; bin/acbuild --debug dep add aci.gonyeo.com/ubuntu
Adding dependency "aci.gonyeo.com/ubuntu"
: trevize acbuild extracttar ; sudo bin/acbuild --debug exec -- ls
Execing: [ls]
meta tag not found on aci.gonyeo.com/ubuntu: expected a 200 OK got 404
Downloading aci.gonyeo.com/ubuntu: [===========================] 197 MB/197 MB 
Uncompressing aci.gonyeo.com/ubuntu: [=========================] 197 MB/197 MB 
There are 12786 entries in the fileMap
Expanding aci.gonyeo.com/ubuntu: [=============================] 197 MB/197 MB 
bin  boot  dev	etc  home  lib	lib64  media  mnt  opt	proc  root  run  sbin  srv  sys  tmp  usr  var
: trevize acbuild extracttar ; 

```

PR #32, opened 2015-10-13
util: UnTar function rewritten to use github.com/coreos/rkt/pkg/tar

dgonyeo:
The UnTar function failed to handle many edge cases in tar files (like
device nodes), so instead of implementing an untar function in acbuild,
a function from rkt is used.


dgonyeo:
Fixes https://github.com/appc/acbuild/issues/28

jonboulle:
LGTM

Issue #19, opened 2015-10-07
*: context free mode


dgonyeo:
A context free mode should be added to support quick modifications to local ACI files.

jonboulle:
Fixed by #34

PR #34, opened 2015-10-13
*: added context free mode

dgonyeo:
The `--modify=path/to/app.aci` flag has been added, and when provided acbuild will transparently wrap the call in `begin` and `end`, with the work context being set to a temp directory for the invocation.

As an example:
```
wget https://aci.gonyeo.com/nginx-latest-linux-amd64.aci
acbuild --modify=nginx-linux-latest-amd64.aci annotation add author "Derek Gonyeo"
```

(note this example won't work at the time of writing due to bug with acbuild's UnTar function)

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,14 +172,15 @@ dependencies.
 
 `acbuild run` also requires root.
 
+### Context-Free Mode
+
+Calling `begin` and `end` with acbuild to make a single change to an existing
+ACI can be cumbersome, so acbuild provides a context free mode. The
+`--modify=path/to/app.aci` flag can be used to specify an ACI to modify, and
+when provided the current build context will be ignored and the change will be
+applied to the given ACI instead.
+
 ## Planned features
-
-### Context-free mode
-
-There are scenarios in which it is not convenient to need to call `begin` and
-`end`, the most obvious being when a single change is made to an existing ACI.
-A flag will be added to allow every subcommand to be performed on a given ACI,
-instead of looking for a current build in progress.
 
 ### Image signing
 
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
 	"strings"
@@ -27,6 +28,7 @@ import (
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/pflag"
 
+	"github.com/appc/acbuild/lib"
 	"github.com/appc/acbuild/util"
 )
 
@@ -53,6 +55,7 @@ var (
 
 	debug       bool
 	contextpath string
+	aciToModify string
 
 	tabOut      *tabwriter.Writer
 	cmdExitCode int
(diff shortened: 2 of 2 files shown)


dgonyeo:
Addresses https://github.com/appc/acbuild/issues/19

acbuild/acbuild.go
```diff
@@ -140,7 +144,45 @@ GLOBAL OPTIONS:
 // terminator.
 func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd *cobra.Command, args []string) {
 	return func(cmd *cobra.Command, args []string) {
+		if aciToModify != "" {
+			contextualCommands := []string{"begin", "end", "abort"}
+			for _, cc := range contextualCommands {
+				command := strings.Split(cmd.Use, " ")[0]
```
dgonyeo: Before anyone can complain about a `slice[0]` without checking its length: https://play.golang.org/p/fTnzbPzmrF

jonboulle: is that a thinly veiled dig at me? ;-)
I don't know why I keep un-learning the slice semantics, too many false memories..

README.md
```diff
@@ -172,14 +172,15 @@ dependencies.
 
 `acbuild run` also requires root.
 
-## Planned features
+### Context Free Mode
 
-### Context-free mode
```
jonboulle: please revert, it should be hyphenated

acbuild/acbuild.go
```diff
@@ -140,7 +144,45 @@ GLOBAL OPTIONS:
...
+					stderr("Can't use the --modify flag with %s.", command)
+					cmdExitCode = 1
+					return
+				}
+			}
+
+			var err error
+			contextpath, err = ioutil.TempDir("", "acbuild-")
+			if err != nil {
+				stderr("%v", err)
```
jonboulle: does stderr(err) not work?

acbuild/acbuild.go
jonboulle: do this outside the loop

acbuild/acbuild.go
```diff
@@ -140,7 +144,45 @@ GLOBAL OPTIONS:
 // terminator.
 func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd *cobra.Command, args []string) {
 	return func(cmd *cobra.Command, args []string) {
+		if aciToModify != "" {
```
jonboulle: might be nicer to invert the conditional 
```
if aciToModify == "" {
 		cmdExitCode = cf(cmd, args)
}
// all aciToModify stuff here
```

acbuild/acbuild.go
dgonyeo: It's not strange slice semantics here, more just knowing that strings.Split will never return an empty slice. The comments there because if I were in your position and noticed it I'd either comment on it or check Split's implementation.

jonboulle:
lgtm

acbuild/acbuild.go
dgonyeo: Nope, the first argument to stderr must be a string.

https://play.golang.org/p/2CvjwRqRwF

jonboulle: derr, err.Error() is what I should have said. nbd.

PR #38, opened 2015-10-15
tests: Added a test for begin

dgonyeo:
Calls begin, and performs a handful of checks to verify that begin creates an empty container.

I want this to get merged before I write more tests, to confirm that I'm going about this the correct way.

The `ACBUILD_BIN` must be set and point to a binary of acbuild for the test to pass.

### DIFF
--- a/tests/begin_test.go
+++ b/tests/begin_test.go
@@ -1,0 +1,72 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"io/ioutil"
+	"path"
+	"testing"
+
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/aci"
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema"
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema/types"
+	"github.com/kylelemons/godebug/pretty"
+)
+
+func TestBegin(t *testing.T) {
+	// Call begin
+	tmpdir, err := setUpTest()
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
+	defer cleanUpTest(tmpdir)
+
+	// Check that there are no files in the rootfs
+	files, err := ioutil.ReadDir(path.Join(tmpdir, ".acbuild", "currentaci", aci.RootfsDir))
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+	if len(files) != 0 {
+		t.Errorf("rootfs in aci contains files, should be empty")
+	}
+
+	// Check that the manifest is no bigger than it needs to be
+	manblob, err := ioutil.ReadFile(path.Join(tmpdir, ".acbuild", "currentaci", aci.ManifestFile))
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+
+	var man schema.ImageManifest
+
+	err = man.UnmarshalJSON(manblob)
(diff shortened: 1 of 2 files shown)


tests/begin_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	man := &schema.ImageManifest{}
+
+	err = man.UnmarshalJSON(manblob)
+	if err != nil {
+		t.Errorf("invalid manifest schema: %v", err)
+	}
+
+	if len(man.Labels) != 0 {
```
jonboulle: General form here is:
```
if g := man.Labels; !reflect.DeepEqual(g, w) {
    t.Errorf("bad labels: got %v, want %v", g, w)
}
```
idea being to test the result more specifically and capture as much useful information as possible in the failure message

dgonyeo: In this case I expect the slice of labels to be empty. The documentation for `reflect.DeepEqual` reads `An empty slice is not equal to a nil slice`. In your example, would my `w` be a nil slice or an empty slice?

All of the following comparisons result in the test failing:
```
if g := man.Labels; !reflect.DeepEqual(g, types.Labels{}) {
if g := man.Labels; !reflect.DeepEqual(g, nil) {
if g := man.Labels; !reflect.DeepEqual(g, make(types.Labels, 0)) {
```

jonboulle: Really? what is fmt.Sprintf("#+v", man.Labels)?

dgonyeo: `fmt.Printf("#%v\n", man.Labels)`
results in:
```
#+v
%!(EXTRA types.Labels=[])
```

dgonyeo: `fmt.Printf("%v\n", man.Labels)` nets me `[]`

jonboulle: Sorry, typo, meant %+v

dgonyeo:
```
    if g := man.Labels; !reflect.DeepEqual(g, make(types.Labels, 0)) {              
        t.Errorf("bad labels: expected no labels but got %+v", g)                   
    }
```

`begin_test.go:55: bad labels: expected no labels but got []`

tests/begin_test.go
```diff
@@ -0,0 +1,74 @@
...
+	if len(man.Dependencies) != 0 {
+		t.Errorf("manifest contains dependencies, shouldn't")
+	}
+	if len(man.PathWhitelist) != 0 {
+		t.Errorf("manifest contains a path whitelist, shouldn't")
+	}
+
+	err = cleanupTest(tmpdir)
+	if err != nil {
+		t.Errorf("%v", err)
```
jonboulle: a) my super grammar nit, setupTest --> setUpTest and cleanupTest --> cleanUpTest
b) defer cleanUpTest() after setUpTest()
c) unless you have a good reason to care about the error that's actually related to acbuild behaviour, then don't log it using t.Errorf; either ignore or panic. any call to Errorf should reflect an issue with the actual lib behaviour, but AFAICT here you're just trying to be a good citizen?

dgonyeo:

tests/begin_test.go
```diff
@@ -0,0 +1,76 @@
...
+		App: &types.App{
+			Exec:  nil,
+			User:  "0",
+			Group: "0",
+		},
+	}
+
+	str := pretty.Compare(man, expectedMan)
+
+	if str != "" {
```
jonboulle: collapse this with 71

tests/begin_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	man := &schema.ImageManifest{}
+
+	err = man.UnmarshalJSON(manblob)
+	if err != nil {
+		t.Errorf("invalid manifest schema: %v", err)
+	}
+
+	acid, err := types.NewACIdentifier("acbuild-unnamed")
```
jonboulle: types.MustACIdentifier

tests/begin_test.go
```diff
@@ -0,0 +1,76 @@
...
+	if len(files) != 0 {
+		t.Errorf("rootfs in aci contains files, should be empty")
+	}
+
+	manblob, err := ioutil.ReadFile(path.Join(tmpdir, ".acbuild", "currentaci", aci.ManifestFile))
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+
+	man := &schema.ImageManifest{}
```
jonboulle: nit, it seems slightly more idiomatic/common to use the `var man schema.ImageManifest` form for objects about to be unmarshaled

jonboulle:
:+1: a few nits but LGTM

PR #39, opened 2015-10-19
Check if overlay fstype is available before loading the module

krobertson:
This change will check if the overlay fstype is already available before
attempting to load the module.

This can help acbuild run better from within a container itself, since often
the kernel modules won't be available to container, but the module will
potentially already be loaded.

@jonboulle @dgonyeo


jonboulle:
:+1: thanks!

Issue #37, opened 2015-10-14
*: deal with versioning


jonboulle:
We need to start thinking about how acbuild works w.r.t versions of the spec.
Initially at least acbuild needs to be versioned, I suggest it could just be 1:1 with the spec itself (i.e. acbuild --version would be equal to the version of appc that is vendored in)
Next would be to start thinking about forwards/backwards compatibility.

PR #41, opened 2015-10-19
godep: update appc/spec dependency

dgonyeo:
acbuild was using an older version of the appc/spec, which was causing
an issue with GetACI in registry/registry.go when the exec command of
a dependency wasn't set.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -7,33 +7,33 @@
 	"Deps": [
 		{
 			"ImportPath": "github.com/appc/spec/aci",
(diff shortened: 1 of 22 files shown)


jonboulle:
Why isn't godeps.json updated?

jonboulle:
LGTM

jonboulle:
@dgonyeo can you please look at #37 soon?

dgonyeo:
Yup, was planning on asking you about the best way to go about that in person tomorrow.

Issue #17, opened 2015-10-07
begin: image fetching


dgonyeo:
`begin` should support being given an image name instead of a local file as a starting point. acbuild would then download the ACI and use it as the starting point for the new build.

```
acbuild begin quay.io/coreos/alpine-sh
```

jonboulle:
Related: #8

jonboulle:
This is really useful but the exact UX for specifying images becomes complex quickly (coreos/rkt#715). 

For now can we leverage what we do in rkt for this? (discovery, local file, http URL, docker://)

jonboulle:
Workaround right now is using something out-of-band to do discovery and retrieve the ACI manually.

PR #43, opened 2015-10-19
begin: add support for fetching remote ACIs to start from

dgonyeo:
With this change if the ACI provided to begin does not exist on the
filesystem, meta discovery is performed to locate and download the ACI.

### DIFF
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -143,8 +143,25 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd 
 		}
 		defer os.Remove(buildpath)
 
+		finfo, err := os.Stat(aciToModify)
+		switch {
+		case os.IsNotExist(err):
+			stderr("ACI doesn't appear to exist: %s.", aciToModify)
+			cmdExitCode = 1
+			return
+		case err != nil:
+			stderr("Error accessing ACI to modify: %v.", err)
+			cmdExitCode = 1
+			return
+		case finfo.IsDir():
+			stderr("ACI to modify is a directory: %s.", aciToModify)
+			cmdExitCode = 1
+			return
+		}
+
 		a := lib.NewACBuild(buildpath, debug)
-		err = a.Begin(aciToModify)
+
+		err = a.Begin(aciToModify, false)
 		if err != nil {
 			stderr("%v", err)
 			cmdExitCode = 1
--- a/lib/begin.go
+++ b/lib/begin.go
@@ -16,12 +16,16 @@ package lib
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
+	"strings"
 
+	"github.com/appc/acbuild/registry"
 	"github.com/appc/acbuild/util"
 
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/aci"
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/discovery"
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema"
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema/types"
 )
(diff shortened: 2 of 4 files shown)


dgonyeo:
Addresses https://github.com/appc/acbuild/issues/17

acbuild/acbuild.go
```diff
@@ -143,8 +143,26 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
 		}
 		defer os.Remove(buildpath)
 
+		finfo, err := os.Stat(aciToModify)
+		if err != nil {
```
jonboulle: how about
```
switch {
case os.IsNotExist(err):
case err != nil:
case finfo.IsDir():
}
```

lib/begin.go
```diff
@@ -112,3 +112,70 @@ func (a *ACBuild) Begin(start string) (err error) {
 
 	return nil
 }
+
+func (a *ACBuild) beginWithStart(start string, insecure bool) error {
```
jonboulle: `beginWithStart`?! has to be a better name. beginFromImage perhaps?

registry/fetch.go
```diff
@@ -50,16 +50,25 @@ func (r Registry) tmpuncompressedpath() string {
 	return path.Join(r.DepStoreTarPath, "tmp.uncompressed.aci")
 }
 
-// FetchAndRender will fetch the given image and all of its dependencies if
-// they have not been fetched yet, and will then render them on to the
-// filesystem if they have not been rendered yet.
-func (r Registry) FetchAndRender(imagename types.ACIdentifier, labels types.Labels, size uint) error {
+func (r Registry) Fetch(imagename types.ACIdentifier, labels types.Labels, size uint, fetchDeps bool) error {
```
jonboulle: // Fetch will foo a bar

registry/fetch.go
```diff
@@ -50,16 +50,25 @@ func (r Registry) tmpuncompressedpath() string {
...
+	return err
+}
+
+// FetchAndRender will fetch the given image and all of its dependencies if
+// they have not been fetched yet, and will then render them on to the
+// filesystem if they have not been rendered yet.
+func (r Registry) FetchAndRender(imagename types.ACIdentifier, labels types.Labels, size uint) error {
+	err := r.Fetch(imagename, labels, size, true)
+	if err != nil {
+		return nil
```
jonboulle: err?

lib/begin.go
```diff
@@ -112,3 +112,70 @@ func (a *ACBuild) Begin(start string) (err error) {
...
+		return err
+	}
+
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		if strings.HasPrefix(file.Name(), "sha512-") {
```
jonboulle: This seems very leaky with the registry, surely we can do better.

lib/begin.go
```diff
@@ -112,3 +112,70 @@ func (a *ACBuild) Begin(start string) (err error) {
...
+			return util.UnTar(path.Join(dir, file.Name()), a.CurrentACIPath, nil)
+		}
+	}
+
+	return fmt.Errorf("aci missing after download?")
```
jonboulle: Probably panic here..?

dgonyeo:

lib/begin.go
```diff
@@ -112,3 +112,70 @@ func (a *ACBuild) Begin(start string) (err error) {
...
+
+	files, err := ioutil.ReadDir(tmpDepStoreTarPath)
+	if err != nil {
+		return err
+	}
+
+	if len(files) == 0 {
+		panic("ACI is missing from temporary dependency store after download")
+	} else if len(files) > 1 {
+		panic("too many files in temporary dependency store after download")
```
jonboulle: print the files? :-)

jonboulle: how about just != 0

jonboulle:
Assuming it works, code lgtm :-)

acbuild/acbuild.go
```diff
@@ -143,8 +143,25 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
...
+			return
+		case finfo.IsDir():
+			stderr("ACI to modify is a directory: %s.", aciToModify)
+			cmdExitCode = 1
+			return
+		}
+
 		a := lib.NewACBuild(buildpath, debug)
-		err = a.Begin(aciToModify)
+
```
jonboulle: sometimes I feel like you randomly insert empty lines just to troll or see if I notice..

dgonyeo: often after looking at something I get this feeling like it'd be easier to read if there was more whitespace, and so I put it in :p

PR #44, opened 2015-10-20
*: refactored library functions, added struct with common fields

dgonyeo:
The library was refactored so that all the exposed functions were
attached to a struct containing the information about the current
context. This makes using the library much cleaner.

The Registry struct had some of its fields renamed to be consistent with
the new struct in the library.

### DIFF
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -18,9 +18,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"strings"
-	"syscall"
 	"text/tabwriter"
 	"text/template"
 
@@ -29,7 +27,6 @@ import (
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/pflag"
 
 	"github.com/appc/acbuild/lib"
-	"github.com/appc/acbuild/util"
 )
 
 const (
@@ -38,8 +35,6 @@ const (
 )
 
 var (
-	workprefix = ".acbuild"
-
 	commandUsageTemplate *template.Template
 
 	templFuncs = template.FuncMap{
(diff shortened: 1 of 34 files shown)


dgonyeo:
My motivation for doing this is that I was pretty unhappy with the code in `acbuild/` related to actually using the acbuild library. This refactor makes the calling of all the functions easier, and makes it so that I don't have five different functions with different naming schemes around to generate different relevant paths.

lib/set-exec.go
```diff
@@ -22,12 +22,12 @@ import (
 )
 
 // SetExec sets the exec command for the untarred ACI stored at acipath.
```
jonboulle: acipath, here and elsewhere, could be updated..

lib/run.go
```diff
@@ -220,7 +223,8 @@ func genDeplist(acipath string, reg registry.Registry) ([]string, error) {
 			return nil, err
 		}
 
-		subdeps, err := genDeplist(path.Join(reg.Scratchpath, depkey), reg)
+		subdeps, err := genDeplist(
```
jonboulle: Eh, I really don't care for the line wrapping, typically in Go we only wrap comments or some egregiously long lines

lib/common.go
```diff
@@ -0,0 +1,25 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lib
+
+type ACBuild struct {
```
jonboulle: // ACBuild foos a bar...

acbuild/acbuild.go
```diff
@@ -61,30 +61,22 @@ var (
...
-
-func workpath() string {
-	return path.Join(contextpath, workprefix, "work")
-}
-
 func lockpath() string {
 	return path.Join(contextpath, workprefix, "lock")
 }
 
+func newACBuild() *lib.ACBuild {
```
jonboulle: Can you pass contextpath and workprefix to newACBuild? global vars always make me uneasy, especially when you are mutating them at times outside of flag parsing.

jonboulle:
A few comments, but this is a good step forward

lib/run.go
dgonyeo: Kk, I'll stop trying to keep everything to <= 80 characters.

dgonyeo:
Did the things you wanted, and also moved all code related to acquiring/releasing the file lock into the `lib/` package. Another pass for the new changes would be good.

lib/set-user.go
```diff
@@ -24,8 +24,19 @@ import (
 )
 
 // SetUser sets the user the pod will run as in the untarred ACI stored at
-// acipath.
-func SetUser(acipath, user string) error {
+// a.CurrentACIPath.
+func (a *ACBuild) SetUser(user string) (err error) {
+	err = a.getLock()
```
jonboulle: Two nits:
1) now that it's no longer returning anything, s/getLock/lock/g and s/releaseLock/unlock/g
2) abbreviate these constructions everywhere:
```
if err = a.lock(); err != nil {
    return
}
defer func() {
  if err1 := a.releaseLock(); err1 != nil {
    err = err1
}()
```

jonboulle:
One more comment, but LGTM.

PR #45, opened 2015-10-21
version: added command

dgonyeo:
The version command was added. The version string gets set at compile
time by linker flags. If the build script is not used, the version will
read: "0.0.0+was-not-built-correctly".

Note that the build now fails when there are no git tags.

### DIFF
--- a/acbuild/version.go
+++ b/acbuild/version.go
@@ -1,0 +1,46 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
+
+	"github.com/appc/acbuild/lib"
+)
+
+var (
+	cmdVersion = &cobra.Command{
+		Use:     "version",
+		Short:   "Get the version of acbuild",
+		Example: "acbuild version",
+		Run:     runWrapper(runVersion),
+	}
+)
+
+func init() {
+	cmdAcbuild.AddCommand(cmdVersion)
+}
+
+func runVersion(cmd *cobra.Command, args []string) (exit int) {
+	if len(args) != 0 {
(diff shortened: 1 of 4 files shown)


dgonyeo:
Addresses https://github.com/appc/acbuild/issues/37

dgonyeo:
As far as what values to use for the tags of acbuild, I think just having them track the version of the appc spec acbuild uses would be appropriate.

lib/version.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lib
+
+var Version = "0.0.0+was-not-built-correctly"
```
jonboulle: Hm, didn't we discuss just having this pinned to github.com/appc/spec/version.Version + git?

dgonyeo: Turns out github.com/appc/spec/schema.version is unexported:
https://github.com/appc/spec/blob/master/schema/version.go

jonboulle: Use AppContainerVersion. c.f. https://github.com/coreos/rkt/blob/9ddb77f0fbc8d5cabbd5a70b7873882341d87ce7/rkt/version.go#L38

jonboulle: Maybe we should just take the rkt approach and print both. It's unrealistic to expect them to be in complete lockstep.

dgonyeo: Oh wow, I completely missed that. Will do.

dgonyeo: SGTM

dgonyeo:

lib/version.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lib
+
+import "github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema"
+
+var Version = "0.0.0+was-not-built-correctly"
+var AppcVersion = schema.AppContainerVersion
```
jonboulle: const?

acbuild/version.go
```diff
@@ -0,0 +1,46 @@
...
+
+func runVersion(cmd *cobra.Command, args []string) (exit int) {
+	if len(args) != 0 {
+		cmd.Usage()
+		return 1
+	}
+
+	stdout("acbuild version %s", lib.Version)
+	stdout("appc version %s", lib.AppcVersion)
+
```
jonboulle: newlines newlines everywhere and nothing to drink

jonboulle:
ship ittttt

lib/version.go
dgonyeo: nope

PR #46, opened 2015-10-21
write: Update to perform overwriting with truncate rather than remove

krobertson:
This changes the overwrite handling to adding the O_TRUNC flag onto the
OpenFile, rather than performing a remove and then recreate.

One primary use for this is running acbuild from a container, where the
ACI file is a bind mount in directly to the file. In this case, the
remove call won't work, but truncating the file is possible.

@jonboulle @dgonyeo


dgonyeo:
LGTM

PR #47, opened 2015-10-21
write: Instrumented gzip compression on the resulting image

krobertson:
This changes the write operation to compress the resulting image with gzip.
    
Currently, it only does gzip. The ACI spec references gzip, bzip2, and xz.
Go has a built in gzip writer, but only a bzip2 reader. XZ is possible, however
it requires linking to libxz and then requiring it at runtime.
    
This changes also reorders some of the file closing within the write function.
It moves the closes from being done in line to being defered. The gzip.Writer
exposes a Close function, and want to ensure the Close options are done in
the proper order, and done across all three.
    
The ACI may be removed on a failure, however it is fine to remove the file
before closing, it will clean up properly.

### DIFF
--- a/acbuild/write.go
+++ b/acbuild/write.go
@@ -33,7 +33,7 @@ var (
 func init() {
 	cmdAcbuild.AddCommand(cmdWrite)
 
-	cmdWrite.Flags().BoolVar(&overwrite, "overwrite", false, "overwrite output ACI")
+	cmdWrite.Flags().BoolVar(&overwrite, "overwrite", false, "overwrite the resulting ACI")
 	cmdWrite.Flags().BoolVar(&sign, "sign", false, "sign the resulting ACI")
 }
 
--- a/lib/write.go
+++ b/lib/write.go
@@ -16,6 +16,7 @@ package lib
 
 import (
 	"archive/tar"
+	"compress/gzip"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


acbuild/acbuild.go
```diff
@@ -153,7 +153,7 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
 
 		cmdExitCode = cf(cmd, args)
 
-		err = a.Write(aciToModify, true, false, nil)
+		err = a.Write(aciToModify, true, false, false, nil)
```
dgonyeo: I actually think it would make more sense for it to compress the image in this case. Thoughts, @jonboulle?

acbuild/write.go
```diff
@@ -33,7 +34,8 @@ var (
 func init() {
 	cmdAcbuild.AddCommand(cmdWrite)
 
-	cmdWrite.Flags().BoolVar(&overwrite, "overwrite", false, "overwrite output ACI")
+	cmdWrite.Flags().BoolVar(&overwrite, "overwrite", false, "overwrite the resulting ACI")
+	cmdWrite.Flags().BoolVar(&compress, "compress", false, "compress the resulting ACI")
```
dgonyeo: Same as earlier I suppose, I think it would make more sense to default to compressing the ACI.

dgonyeo:
Other than my comments on defaulting to compress the ACI, LGTM.

acbuild/acbuild.go
krobertson: I had thought about this but wasn't sure... I would agree with compressing them by default. It seems more ideal.

acbuild/write.go
krobertson: If it compresses by default, then should I make the option `--skip-compression`?

In that case, then you'd either be doing `--compress=false` or `--skip-compression`.

Is not compressing really a need?

jonboulle: how about `--no-compression` like actool?
Otherwise I'm happy to just omit this until we have a use case - at that time it'll take five minutes to add.

dgonyeo: I'm in favor of dropping support for skipping compression.

jonboulle: Ship it

krobertson:
Updated to simply always compress. Removed the parameter and passing of the compress parameter. Give it another look over.

dgonyeo:
LGTM

PR #48, opened 2015-10-21
build: tweak required to build script based on tag github generated

dgonyeo:
Looks like when a tag is made via github's release UI, the tag isn't
annotated. This required a tweak to the build script to prevent the
following error:

fatal: No annotated tags can describe 'ccfe28dd8b55ca76995b72cd9d028745e2d98d14'.
However, there were unannotated tags: try --tags.


jonboulle:
Rather, let's always sign (annotate) tags...

PR #52, opened 2015-10-21
build: removed quotes from acbuild version string in build script

dgonyeo:
Was:
./bin/acbuild version
acbuild version "v0.1.0"
appc version 0.7.1+git

Is now:
./bin/acbuild version
acbuild version v0.1.0-dirty
appc version 0.7.1+git


jonboulle:
Welp.

PR #53, opened 2015-10-22
README: updated to reflect new features

dgonyeo:
Image signing and image fetching for begin were both still listed as
planned features, but they have been implemented.

Also added [cnt](https://github.com/blablacar/cnt) to the related works section.


jonboulle:
lgtm

PR #54, opened 2015-10-22
tests: added tests for annotation add/rm

dgonyeo:
Handful of tests were added for adding and removing annotations. Issue
was found where an error wouldn't be returned if you asked acbuild to
remove a nonexistent annotation/label/mount/etc, this commit also fixes
that.

Also added a test script. Needs to be run as root for all tests to pass.


dgonyeo:
Closing in favor of a larger PR I'm about to make.

Issue #24, opened 2015-10-07
*: add some initial tests for every command


dgonyeo:
Each command supported by acbuild should have at least one functional test.

jonboulle:
Perhaps you could edit your comment to add a check-list for this one.

jonboulle:
One possibility is to do something like fleet: https://github.com/coreos/fleet/blob/master/functional/util/util.go#L29
https://github.com/coreos/fleet/blob/master/functional/README.md

(rkt is similar but rather more complex)

PR #55, opened 2015-10-22
tests: added tests for all commands except run, write, copy, and version

dgonyeo:
An issue was found where remove commands will not emit an error if asked to remove a nonexistent item. This commit fixes that.

Additionally this commit adds test for the following commands:
- annotation add
- annotation remove
- dependency add
- dependency remove
- end
- environment add
- environment remove
- label add
- label remove
- mount add
- mount remove
- port add
- port remove
- set-exec
- set-group
- set-name
- set-user

### DIFF
--- a/lib/common.go
+++ b/lib/common.go
@@ -25,6 +25,10 @@ import (
 
 const defaultWorkPath = ".acbuild"
 
+// ErrNotFound is returned when acbuild is asked to remove an element from
+// a list and the element is not present in the list
+var ErrNotFound = fmt.Errorf("element to be removed does not exist in this ACI")
+
 // ACBuild contains all the information for a current build. Once an ACBuild
 // has been created, the functions available on it will perform different
 // actions in the build, like updating a dependency or writing a finished ACI.
--- a/test
+++ b/test
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+set -e
+
+ORG_PATH="github.com/appc"
+REPO_PATH="${ORG_PATH}/acbuild"
+TESTS_PATH="${REPO_PATH}/tests"
+
+export ACBUILD_BIN="${PWD}/bin/acbuild"
+export GOPATH="${PWD}/gopath"
+
+source "${GOPATH}/src/${REPO_PATH}/build"
+
+go test "${TESTS_PATH}" "$@"
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -73,6 +73,14 @@
 			"Rev": "76626ae9c91c4f2a10f34cad8ce83ea42c93bb75"
 		},
 		{
+			"ImportPath": "github.com/kylelemons/godebug/diff",
+			"Rev": "21cb3784d9bda523911b96719efba02b7e983256"
+		},
+		{
+			"ImportPath": "github.com/kylelemons/godebug/pretty",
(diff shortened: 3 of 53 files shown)


dgonyeo:
Addresses most of https://github.com/appc/acbuild/issues/24

jonboulle:
>  remove commands will not emit an error if asked to remove a nonexistent item. 

SHOULD we emit an error in this case? There might be something to be said for commands being (optionally?) idempotent?

dgonyeo:
I think if you try to remove a mount point, and the mount point doesn't exist, it should at least say something. Perhaps just print a warning to stderr and have a 0 exit code?

dgonyeo:
Honestly I'd just prefer to leave it on the user to ignore the error if they're removing something they're not sure would exist.

jonboulle:
I am somewhat receptive to that but then it should be something consistent at the CLI layer (e.g. probably a particular error code) so that people can actually predictably ignore it.

/cc @krobertson thoughts?

test
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -e
+
+ORG_PATH="github.com/appc"
+REPO_PATH="${ORG_PATH}/acbuild"
+TESTS_PATH="${REPO_PATH}/tests"
+
+export ACBUILD_BIN="${PWD}/bin/acbuild"
+export GOPATH=${PWD}/gopath
```
jonboulle: Quote, for consistency

test
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -e
+
```
jonboulle: source ./build ?

dgonyeo: Sure, but it adds about 2 seconds on my laptop to running the tests. Probably for the best, so I don't confuse myself in the future as to why changes to `lib/` aren't being reflected in the tests.

dgonyeo:
Should be ready for another pass. When acbuild is asked to remove something that doesn't exist the exit code will be 2, so that users can check what type of error was encountered. It'll also be easy going forward to add other error types with different exit codes.

lib/common.go
```diff
@@ -25,6 +25,22 @@ import (
 
 const defaultWorkPath = ".acbuild"
 
+// ErrNotFound is returned when acbuild is asked to remove an element from
+// a list and the element is not present in the list
+var ErrNotFound = fmt.Errorf("element to be removed does not exist in this ACI")
+
+// GetErrorCode will return the recommended exit code for acbuild given an error
+func GetErrorCode(err error) int {
```
jonboulle: Interesting approach. I think this function should exist in the CLI package instead though (but still key off the lib.Err types)

dgonyeo: Yeah, I guess it does make more sense to be there. Fixed.

dgonyeo:

test
krobertson: you can possibly do a `go test -i ./...` before `go test` itself.

krobertson:
LGTM.  Oddly, github doesn't do syntax highlighting on the "set-user_test.go" files... maybe doesn't like the dash?

For the previous point about erroring when removing something not found, general command line behavior is to error.  If you do just `rm foo` and foo doesn't exist, it'll be an error. Maybe a `-f` flag would make sense in the future?

test
dgonyeo: All of the dependencies for the tests should be vendored, so I don't think that would do anything.

Issue #56, opened 2015-10-23
Panic while uncompressing dependency


jcollie:
After updating to the lastest master this morning, I started getting the following error:

    [jcollie@pc28043 rkt-logstash]$ acbuild version
    acbuild version v0.1.0-4-g40f07a0
    appc version 0.7.1+git
    [jcollie@pc28043 rkt-logstash]$ sudo acbuild begin
    [jcollie@pc28043 rkt-logstash]$ sudo acbuild dependency add dmacc.net/centos7 --image-id="sha512-baae275b452242b1ccdba64bff15a7a9" --label="version=0.1"
    [jcollie@pc28043 rkt-logstash]$ sudo acbuild label add os linux
    [jcollie@pc28043 rkt-logstash]$ sudo acbuild label add arch amd64
    [jcollie@pc28043 rkt-logstash]$ sudo acbuild run --debug -- yum -y update
    Running: [yum -y update]
    Downloading dmacc.net/centos7: [===============================] 62.9 MB/62.9 MB
    panic: runtime error: makeslice: len out of range==            ] 41.6 MB/62.9 MB
    
    goroutine 1 [running]:
    strings.Repeat(0x8ca320, 0x1, 0xfffffffffffffff5, 0x0, 0x0)
            /usr/local/go/src/strings/strings.go:464 +0x5f
    github.com/appc/acbuild/Godeps/_workspace/src/github.com/coreos/ioprogress.DrawTextFormatBarForW.func2(0x4f38000, 0x3bfe5e9, 0x0, 0x0)
            /opt/acbuild/gopath/src/github.com/appc/acbuild/Godeps/_workspace/src/github.com/coreos/ioprogress/draw.go:114 +0xcf
    github.com/appc/acbuild/registry.newIoprogress.func1(0x4f38000, 0x3bfe5e9, 0x0, 0x0)
            /opt/acbuild/gopath/src/github.com/appc/acbuild/registry/fetch.go:441 +0x244
    github.com/appc/acbuild/Godeps/_workspace/src/github.com/coreos/ioprogress.DrawTerminalf.func1(0x4f38000, 0x3bfe5e9, 0x0, 0x0)
            /opt/acbuild/gopath/src/github.com/appc/acbuild/Godeps/_workspace/src/github.com/coreos/ioprogress/draw.go:54 +0xf8
    github.com/appc/acbuild/Godeps/_workspace/src/github.com/coreos/ioprogress.(*Reader).drawProgress(0xc820063d10)
[...]

PR #57, opened 2015-10-23
run: --debug flag caused a panic with the Uncompress progress bar

dgonyeo:
The recent refactor resulted in the registry being correctly passed the
debug value, which exposed an issue with an additional progress bar that
was being drawn when uncompressing a dependency.

The panic was a combination of the logic drawing the bar passing the
ioprogress library a number that was too small for the total size, and
the ioprogress library not having a patch to be able to handle incorrect
numbers.

This commit removes the uncompressing bar, and updates the ioprogress
library.

Fixes https://github.com/appc/acbuild/issues/56

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -41,7 +41,7 @@
 		},
 		{
 			"ImportPath": "github.com/coreos/ioprogress",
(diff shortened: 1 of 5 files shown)


jonboulle:
LGTM

PR #58, opened 2015-10-23
begin: fixed a silly typo that broke fetching remote ACIs

dgonyeo:
Sorry about this, but it looks like I typo'd a number when making a change for code review at some point. Don't know how I didn't catch this sooner.


jonboulle:
hardest problems in computer science

PR #61, opened 2015-10-23
examples: added check if build-nginx.sh is running as root

dgonyeo:
build-nginx.sh must be run as root. This adds a check for that.

### DIFF
--- a/examples/build-nginx.sh
+++ b/examples/build-nginx.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 set -e
 
+if [ "$EUID" -ne 0 ]; then
+    echo "This script uses functionality which requires root privileges"
+    exit 1
+fi
+
 acbuild --debug begin quay.io/listhub/alpine
 
 trap "{ export EXT=$?; acbuild --debug end && exit $EXT; }" EXIT


jonboulle:
How about just putting it in the actual script?

dgonyeo:
That makes more sense. How's that look?

examples/build-nginx.sh
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 set -e
 
+if [ "$EUID" -ne 0 ]
```
jonboulle: ]; then

examples/build-nginx.sh
```diff
@@ -1,6 +1,12 @@
...
+    then
+    echo "This script must be run as root."
+    exit
```
jonboulle: exit 1 probably

examples/build-nginx.sh
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 set -e
 
+if [ "$EUID" -ne 0 ]
+    then
+    echo "This script must be run as root."
```
jonboulle: perhaps "This script uses functionality (run) which requires root privileges" or something?

dgonyeo: If it was just `run` I think I'd get away with some scattered `sudo`s, but because `begin` needs to be run as root (for chrooting during the ACI extracting) every single command in here would need sudo to be able to modify the files in `.acbuild`.

dgonyeo: I think "This script uses functionality which requires root privileges" would be fine though

jonboulle:
ship it

PR #64, opened 2015-10-23
vagrant: added initial vagrant scripts

dgonyeo:
Added a Vagrantfile and relevant scripts (largely stolen from rkt).
`vagrant up` should produce a machine with both acbuild and rkt 0.9.0
(will be updated once rkt is released) in the path.


jonboulle:
LGTM if it works on a mac :trollface:

Issue #36, opened 2015-10-14
Issue untar'ing some ACIs


dgonyeo:
Certain ACIs cause the following error when acbuild attempts to untar them:
```
extracttar error: exit status 1, output: error extracting tar: error extracting tarball: archive/tar: invalid tar header
```

An example of an ACI that causes this is at https://aci.gonyeo.com/nginx-latest-linux-amd64.aci. The application was made with buildroot, and the ACI was built with actool. rkt has no issues running the ACI.

ecnahc515:
I'm experiencing this with `docker://debian` as well. Converted w/ docker2aci.

PR #68, opened 2015-10-26
util: handle compressed ACIs

iaguis:
When starting the build from an ACI, it was assumed it was an
uncompressed tar.

This commit uses NewCompressedReader from appc/spec/aci to handle ACIs
compressed with gzip, bzip2 and xz.

Fixes https://github.com/appc/acbuild/issues/36

### DIFF
--- a/util/files.go
+++ b/util/files.go
@@ -15,9 +15,11 @@
 package util
 
 import (
+	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


jonboulle:
Wow, oops. Great catch! Thanks.

PR #69, opened 2015-10-26
*: rename UnTar -> ExtractImage

jonboulle:
Noticed in #68. The latter is more accurate, particularly since this now
deals with compressed images. Also cleans up a few other variable names.


iaguis:
Makes sense. LGTM

PR #70, opened 2015-10-26
examples: if sudo is required, sudo the script instead of exiting

dgonyeo:
I think this is a little more user friendly.


jonboulle:
ehh, -1

jonboulle:
Let's not second guess users. One of the goals with appc is to make security as explicit as possible.

dgonyeo:
got it

ecnahc515:
What about just printing a warning and exiting before doing anything?

dgonyeo:
That's what the script currently does.

jonboulle:
isn't that exactly what we do?

ecnahc515:
Oh, your right. Ignore.

Issue #67, opened 2015-10-25
Concurrent build issues


ecnahc515:
If I'm invoking two acbuilds acbuild where the first is using `begin` and the second is using `--modify`, the second one fails to start because the lock for the first is created in the current directory, and the second cannot acquire the lock.

It would be great if the `.acbuild` context was named after the build, and the lock was associated with a particular `.acbuild` context (or aci if using `--modify`).

Ex:
`acbuild begin path/to/example.aci --name example-app` where --name must be unique for this particular build, to allow for multiple concurrent builds.
This would result in something like:
`.acbuild-example-app` and a lock named `.acbuild-example-app.lock` in the current directory.

`acbuild --modiy path/to/example.aci set name coreos.com/example` could produce a lock named something like:
`.path_to_example.aci.lock`

You could still only have one instance of an `acbuild begin` in a single directory, but you could have as many `acbuild --modify` running in the same directory, assuming each have their own ACI to modify (still limited to one acbuild per aci).

dgonyeo:
I suppose I should create a temp directory for --modify builds

dgonyeo:
Now that I've taken a look at this, the issue is actually sillier than I originally thought. `acbuild` does make a temp directory, which is used by `begin`, `write`, and `end`, but it's not used by the given command you want to run.

PR #71, opened 2015-10-26
*: --modify was broken, didn't call command with different work context

dgonyeo:
Fixes https://github.com/appc/acbuild/issues/67

The --modify flag creates a new temp directory, and then uses that as a
the build context for a call to begin, the given command, write, and
end. It was failing to use the alternate work context for the given
command.

The temp directory's path used by the --modify flag is also now based on
a sha512 hash of the absolute path of the given ACI to modify. This aims
to prevent concurrent invocations of --modify on the same ACI.

### DIFF
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -15,9 +15,11 @@
 package main
 
 import (
+	"crypto/sha512"
 	"fmt"
-	"io/ioutil"
 	"os"
+	"path"
+	"path/filepath"
 	"strings"
 	"text/tabwriter"
 	"text/template"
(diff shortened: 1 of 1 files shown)


acbuild/acbuild.go
```diff
@@ -159,7 +152,29 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
...
+			contextpath = contextpath[:16]
+		}
+
+		err = os.MkdirAll(contextpath, 0755)
+		if err != nil {
+			stderr("%v", err)
+			cmdExitCode = 1
+			return
+		}
+		defer os.Remove(contextpath)
```
jonboulle: RemoveAll?

dgonyeo: fixed

jonboulle:
LGTM

PR #72, opened 2015-10-26
readme: use github http endpoint

aaronlevy:
SSH endpoint is for r/w access. Document using https so anyone can clone.


dgonyeo:
The ssh endpoint can be used for read only access, but iirc it requires you to have an ssh key setup with github, so this will still be applicable to a wider audience than before.

LGTM.

aaronlevy:
Ah, you're right. I was on a blank vm and got denied and just assumed that was because ssh was r/w.

PR #73, opened 2015-10-26
cat-manifest: added implementation

dgonyeo:
This commit adds a command to cat the manifest, either from the current
build or from a given aci.

### DIFF
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -126,14 +126,15 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd 
 			return
 		}
 
-		contextualCommands := []string{"begin", "write", "end", "version"}
 		command := strings.Split(cmd.Use, " ")[0]
-		for _, cc := range contextualCommands {
-			if command == cc {
-				stderr("Can't use the --modify flag with %s.", command)
-				cmdExitCode = 1
-				return
-			}
+		switch command {
+		case "cat-manifest":
+			cmdExitCode = runCatOnACI(aciToModify)
+			return
+		case "begin", "write", "end", "version":
+			stderr("Can't use the --modify flag with %s.", command)
+			cmdExitCode = 1
+			return
 		}
 
 		finfo, err := os.Stat(aciToModify)
--- a/acbuild/cat-manifest.go
+++ b/acbuild/cat-manifest.go
@@ -1,0 +1,65 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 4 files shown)


jonboulle:
Hmm. This feels a little weird in `acbuild`; displaying the manifest of the current build makes sense, but of a given ACI is a bit odd. Perhaps this functionality should just stay in actool?

dgonyeo:
I added the `--aci` flag to prevent people from doing `acbuild --modify=some.aci cat-manifest`, since that would result in expanding the entire ACI to read a few hundred bytes of a file.

I'd be fine with removing the `--aci` flag, but if I do should I allow `cat-manifest` to be used with the `--modify` flag?

jonboulle:
Oh right. Hmm. I think for now, either just disallow it with --modify, or
have that codepath do the same thing you're doing now with lib.CatManifest.
Essentially --modify is --in-place or --act-on whatever.

dgonyeo:
I chose to have `--modify` use the same code path as was being used in `runCat`. How's that look?

jonboulle:
Better, but I really don't like more and more getting shoehorned into runWrapper :-). Can we split that out?

dgonyeo:
Moved to `acbuild/cat-manifest.go`

acbuild/acbuild.go
```diff
@@ -126,14 +126,17 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
...
+		switch command {
+		case "cat-manifest":
+			cmdExitCode = runCatOnACI(aciToModify)
+			return
+		case "begin", "write", "end", "version":
+			stderr("Can't use the --modify flag with %s.", command)
+			cmdExitCode = 1
+			return
+		default:
+			break
```
jonboulle: Remove this.

acbuild/cat-manifest.go
```diff
@@ -0,0 +1,67 @@
...
+
+func runCatOnACI(aciToModify string) (exit int) {
+	if len(args) != 0 {
+		cmdCat.Usage()
+		return 1
+	}
+	err := lib.CatManifest(aciToModify, prettyPrint)
+	if err != nil {
+		stderr("cat-manifest: %v", err)
+		cmdExitCode = 1
```
jonboulle: Nack, this should be returned, not set...

util/manifest.go
```diff
@@ -75,3 +77,23 @@ func ModifyManifest(fn func(*schema.ImageManifest), acipath string) error {
 
 	return nil
 }
+
+// PrintManifest will print the given manifest to stdout, optionally inserting
+// whitespace to make it more human readable.
+func PrintManifest(man *schema.ImageManifest, prettyPrint bool) error {
+	if !prettyPrint {
```
jonboulle: var manblob, err and then you can share code for most of these paths

dgonyeo:
Should be ready for another pass

lib/cat-manifest.go
```diff
@@ -0,0 +1,95 @@
...
+
+	tr, err := aci.NewCompressedTarReader(file)
+	if err != nil {
+		return fmt.Errorf("error decompressing image: %v", err)
+	}
+	defer tr.Close()
+
+	for {
+		hdr, err := tr.Next()
+		switch {
```
jonboulle: if err != nil but != io.EOF could we get stuck in here?

jonboulle:
Nice, one question, lgtm after that

lib/cat-manifest.go
dgonyeo: I suppose we could.

Added a 
```
case err != nil:
    return err
```

PR #74, opened 2015-10-26
replace-manifest: added command to replace manifest with a new one

dgonyeo:
`replace-manifest` will overwrite the manifest in the current ACI with a given new ACI. Analogous to `actool`'s `patch-manifest --manifest=thing`

### DIFF
--- a/acbuild/replace-manifest.go
+++ b/acbuild/replace-manifest.go
@@ -1,0 +1,56 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
+)
+
+var (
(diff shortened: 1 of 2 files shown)


ecnahc515:
Maybe `set-manifest`?

dgonyeo:
That would be fine with me. I went with `new-manifest` to differentiate it from all the `set-*` commands, since it's not just altering a value in the manifest.

ecnahc515:
Yeah, but it's really an 'overwrite' operation, so I like the word `set`. I also feel similarly about the `add` commands after using it this weekend, but that could be discussed more later.

dgonyeo:
The `add` commands usually append to a list, unless you're adding something that already exists.

jonboulle:
replace-manifest ?

dgonyeo:
I like `replace-manifest` more. New commit pushed.

ecnahc515:
Does `replace` make sense when you have an ACI that doesn't already contain a manifest?

dgonyeo:
I don't know if there ever would be such a situation. Even when you create a minimal ACI with acbuild begin it has a manifest.

jonboulle:
Right, aiui replace is always appropriate.

jonboulle:
Hm. I wonder if the lib function should operate on a file or an imagemanifest object.

dgonyeo:
I *think* if it took an ImageManifest that it would be the only place a user of the library needs to pull in something from the appc/spec repo. I might be wrong on that though, not at my laptop.

ecnahc515:
@dgonyeo I'm thinking a case where we have a rootfs in a tarball but without the manifest, ie: someone made their own tarball but want to use acbuild to create a manifest

dgonyeo:
That's not a valid ACI then. In that case I'd suggest the person untar the rootfs, acbuild begin, and acbuild copy rootfs /

jonboulle:
LGTM but please rename the PR

PR #80, opened 2015-10-28
set-working-directory: added command to set the ACI's working directory

dgonyeo:
Fixes https://github.com/appc/acbuild/issues/78

### DIFF
--- a/acbuild/set-working-dir.go
+++ b/acbuild/set-working-dir.go
@@ -1,0 +1,58 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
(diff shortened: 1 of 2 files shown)


dgonyeo:
I was uncertain how to go about unsetting an ACI's working directory, right now the command to do it is:

```
acbuild set-wd ""
```

I don't think that's as intuitive as it could be. Maybe instead the syntax should be:

```
acbuild wd set /root
acbuild wd unset
```

This would also apply to set-exec, since that's another field that is allowed to not be set.

Thoughts?

jonboulle:
I have a dusty memory of @vcaputo having an opinion on this from a previous discussion sometime...

ecnahc515:
Isn't the default working directory `/` usually? I know that's the case in docker. If it is, then the 'unset' is just `acbuild wd set /`.

dgonyeo:
Not that it would impact the user, but there is a difference in the manifest between the value not existing and the value being "/". Should acbuild differentiate between the two?

ecnahc515:
The spec says it's optional, but the default is `/` but it's overridable by the ACE, so if you always set it, then it's possible the ACE wont override it with it's own custom default. It seems weird that the ACE can change the default at all, so I personally think it should always default to `/`.

dgonyeo:
I'm still going to want to support unsetting the exec field. Maybe just a flag would be sufficient? Something like one of these:
```
acbuild set-exec --unset
acbuild set-exec --empty
acbuild set-exec --reset
```
And if the flag exists there I'd be inclined to add it to `set-wd`. I'm not opposed to defaulting it to `/`, but I still want to provide a way to unset it.

dgonyeo:
Unsetting it is more of an edge case that I can worry about in a future PR. Thoughts on merging this @jonboulle ?

jonboulle:
Sure, let's cross that bridge when we come to it.

PR #81, opened 2015-10-28
set-eh: added commands to set event handlers in the manifest

dgonyeo:
Fixes https://github.com/appc/acbuild/issues/77.

### DIFF
--- a/acbuild/set-event-handlers.go
+++ b/acbuild/set-event-handlers.go
@@ -1,0 +1,85 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


jonboulle:
Hrm, I wonder if there's something we can do here to future-proof https://github.com/appc/spec/issues/58 /  https://github.com/coreos/rkt/issues/1495

dgonyeo:
I imagine if the spec gets changed that more fields would just be added to the event handler object. The tweak that would happen here to accommodate those could just be requiring more arguments get passed in to the command, so something like:

```
acbuild set-eh pre-start root:root /root/setup-stuff.sh
```

Changes like that would be fairly easy with the current implementation. If we want to allow for more complex changes (like changing the uid for an existing event handler) another option would be to add subcommands onto `pre-fetch` and `post-stop`, but those changes would require more rewriting of this.

jonboulle:
OK. I guess let's just roll forward and we can cross that bridge. LGTM.

Issue #42, opened 2015-10-19
Add support for manipulating isolators


dgonyeo:
acbuild should have functionality for adding/removing/modifying isolators in an ACI.

PR #82, opened 2015-10-29
isolators: added commands to add and remove isolators

dgonyeo:
Fixes https://github.com/appc/acbuild/issues/42.

Example commands:
```
acbuild isolator add resource/cpu value.json
echo '{"request":"1G","limit":"2G"}' | acbuild isolator add resource/memory -
```

### DIFF
--- a/acbuild/isolator.go
+++ b/acbuild/isolator.go
@@ -1,0 +1,119 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


acbuild/isolator.go
```diff
@@ -0,0 +1,114 @@
...
+		return 1
+	}
+	if len(args) != 2 {
+		stderr("isolator add: incorrect number of arguments")
+		return 1
+	}
+
+	var val []byte
+	var err error
+	if args[1] == "-" {
```
jonboulle: This is kind of a nit, but my preference for structuring this sort of thing is to switch on the interface rather than all the logic
```
var r io.Reader
var err error
if args[1] == "-" {
  r = os.Stdin
} else {
  r, err = os.Open(args[1])
  ...
}
```

jonboulle:
:+1:

PR #83, opened 2015-10-29
godep: updated version of rkt used to latest (0.10.0+git)

dgonyeo:
Bumping rkt so that acbuild will have the fix to fileutil (https://github.com/coreos/rkt/pull/1689) and the ExtractTarInsecure function (https://github.com/coreos/rkt/pull/1616).


jonboulle:
lgtm

Issue #79, opened 2015-10-28
Add ability to set path whitelist


dgonyeo:

dgonyeo:
Will be blocking until https://github.com/appc/spec/issues/539 either gets shot down or implemented/merged.

Issue #21, opened 2015-10-07
*: add explicit layered mode


dgonyeo:
I think a command like `acbuild layer` should be supported that would add a new layer into the current ACI.

```
acbuild exec -- wget http://example.com/very-big-file.deb
acbuild exec -- dpkg -i ./very-big-file.deb
acbuild exec -- rm ./very-big-file.deb
acbuild layer
```

The above would produce a layered image, but the file `very-big-file.deb` would not exist in any of the layers.

jonboulle:
Going to wait until we sort out push before dealing with this.
#16

PR #84, opened 2015-10-29
path-whitelist: added commands for manipulating path whitelist

dgonyeo:
Commands added:
- add: adds a set of paths to the whitelist
- rm: removes a set of paths from the whitelist
- clear: empties the whitelist

Fixes https://github.com/appc/acbuild/issues/79.


jonboulle:
Isn't path whitelist something we can/should automatically generate, e..g https://github.com/appc/acbuild/issues/21

dgonyeo:
Talked with @jonboulle OOB, closing this for now so I can spend time investigating automatically generating this.

PR #87, opened 2015-10-30
travis: added .travis.yml

dgonyeo:
Added a `.travis.yml` file to allow for travis CI to run.

Required a modification to `tests/common.go` to work in the travis environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: go
+matrix:
+  include:
+    - go: 1.4.2
+    - go: 1.5.1
+
+script:
+ - ./build
+ - ./test
+
+notifications:
+  email: false
+
+sudo: true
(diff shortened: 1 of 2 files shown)


dgonyeo:
This should be good to go now. The `.travis.yml` file needs to call `./build` before `./test` for some reason, but I'd prefer to fix that in a future PR.

jonboulle:
can you explain the tests/common change for posterity?

LGTM

dgonyeo:
Without the tests/common change the travis test [would panic](https://travis-ci.org/appc/acbuild/jobs/88460616) with the following error:
```
panic: read |0: bad file descriptor
```

I assume that the fact that I was reading stdout and stderr in goroutines, and not starting the command, getting the outputs, and then waiting on it, was introducing a race condition, that the travis tests here simply exposed.

PR #89, opened 2015-11-02
util: extracting ACIs no longer requires root

dgonyeo:
Modified the ExtractImage function in the util package to use the new
ExtractTarInsecure function when not running as root.

Also added a check to run for the current user to be root, so that a
nicer error message can be printed.


jonboulle:
LGTM

PR #90, opened 2015-11-02
*: many places were generating an invalid App struct

dgonyeo:
When presented with a manifest that lacked an App, many places in the
lib package would fill it in with a `&types.App{}`. This is incorrect, as
the user and group fields must be set in the struct for the manifest to
be valid. Without this fix, acbuild will error when asked to modify an
existing ACI in such a way that requires creating an App in the
manifest.

### DIFF
--- a/lib/common.go
+++ b/lib/common.go
@@ -20,14 +20,26 @@ import (
 	"path"
 	"syscall"
 
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema/types"
+
 	"github.com/appc/acbuild/util"
 )
 
 const defaultWorkPath = ".acbuild"
 
-// ErrNotFound is returned when acbuild is asked to remove an element from
-// a list and the element is not present in the list
+// ErrNotFound is returned when acbuild is asked to remove an element from a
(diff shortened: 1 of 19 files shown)


jonboulle:
Can you use a function instead? see e.g. historical discussion in https://github.com/appc/spec/pull/124

dgonyeo:
done

lib/common.go
```diff
@@ -20,15 +20,24 @@ import (
...
 
 const defaultWorkPath = ".acbuild"
 
-// ErrNotFound is returned when acbuild is asked to remove an element from
-// a list and the element is not present in the list
+// ErrNotFound is returned when acbuild is asked to remove an element from a
+// list and the element is not present in the list
 var ErrNotFound = fmt.Errorf("element to be removed does not exist in this ACI")
 
+func newManifestApp() *types.App {
```
jonboulle: perhaps a very simple docstring to explicate why this exists

jonboulle:
LGTM. Commit message header should summarise the change rather than the symptom: perhaps "fix invalid uses of empty app struct"

PR #91, opened 2015-11-03
run: stop following symlinks when searching for binary in path

dgonyeo:
When run was searching the path for a given binary, it would attempt to
follow the symlink. This is incorrect, as the symlink might not be valid
until after the chroot. This commit fixes that.

### DIFF
--- a/lib/run.go
+++ b/lib/run.go
@@ -166,18 +166,19 @@ func supportsOverlay() bool {
 }
 
 func findCmdInPath(pathlist []string, cmd, prefix string) (string, error) {
-	if len(cmd) != 0 && cmd[0] == '/' {
+	if path.IsAbs(cmd) {
 		return cmd, nil
 	}
 
 	for _, p := range pathlist {
-		ex, err := util.Exists(path.Join(prefix, p, cmd))
-		if err != nil {
+		_, err := os.Lstat(path.Join(prefix, p, cmd))
+		switch {
+		case os.IsNotExist(err):
+			continue
+		case err != nil:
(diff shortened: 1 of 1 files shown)


lib/run.go
```diff
@@ -171,13 +171,14 @@ func findCmdInPath(pathlist []string, cmd, prefix string) (string, error) {
 	}
 
 	for _, p := range pathlist {
-		ex, err := util.Exists(path.Join(prefix, p, cmd))
```
jonboulle: Should this be a util.Lexists or something?

dgonyeo: I'm actually thinking about just removing the `util.Exists` function. It doesn't really save much code anywhere, since each time I use it I still need to specify what to do if the file doesn't exist, what to do if there was an error, and what to do if the file exists. It doesn't simplify any of the logic.

jonboulle: SGTM

lib/run.go
```diff
@@ -171,13 +171,14 @@ func findCmdInPath(pathlist []string, cmd, prefix string) (string, error) {
...
-		if err != nil {
+		_, err := os.Lstat(path.Join(prefix, p, cmd))
```
yifan-gu: Can we use [path.IsAbs()](https://golang.org/pkg/path/#IsAbs) for `if len(cmd) != 0 && cmd[0] == '/'` ?
@dgonyeo

dgonyeo: Most definitely.

yifan-gu:
LGTM

PR #92, opened 2015-11-03
tests: added tests for cat-man, replace-man, set-eh, and set-wd

dgonyeo:
More work towards https://github.com/appc/acbuild/issues/24.

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -15,8 +15,8 @@
 package tests
 
 import (
+	"bytes"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -44,25 +44,6 @@ var (
 			runtime.GOOS,
 		},
 	}
-
-	emptyManifest = schema.ImageManifest{
-		ACKind:    schema.ImageManifestKind,
-		ACVersion: schema.AppContainerVersion,
-		Name:      *types.MustACIdentifier("acbuild-unnamed"),
-		Labels:    systemLabels,
-	}
-
-	emptyManifestWithApp = schema.ImageManifest{
-		ACKind:    schema.ImageManifestKind,
-		ACVersion: schema.AppContainerVersion,
-		Name:      *types.MustACIdentifier("acbuild-unnamed"),
-		App: &types.App{
-			Exec:  nil,
-			User:  "0",
-			Group: "0",
-		},
-		Labels: systemLabels,
-	}
 )
 
 func init() {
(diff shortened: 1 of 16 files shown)


tests/common.go
```diff
@@ -76,18 +120,7 @@ func init() {
...
-	stdout   []byte
-	stderr   []byte
-}
-
-func (ae acbuildError) Error() string {
-	return fmt.Sprintf("non-zero exit code of %d: %v\nstdout:\n%s\nstderr:\n%s", ae.exitCode, ae.err, string(ae.stdout), string(ae.stderr))
-}
-
-func runACBuild(workingDir string, args ...string) *acbuildError {
+func runACBuild(workingDir string, args ...string) (int, []byte, []byte, error) {
```
yifan-gu: Want to comment what's being returned?

tests/replace-manifest_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"encoding/json"
+	"os"
+	"testing"
+)
+
+func TestReplaceManifest(t *testing.T) {
+	workingDir := setUpTest(t)
+	defer cleanUpTest(workingDir)
+
+	manblob, err := detailedManifest.MarshalJSON()
```
yifan-gu: nit: use json.Marshal() to keep consistency?

dgonyeo: `schema.ImageManifest.MarshalJSON` calls `assertValid` for me though, to make sure that there isn't something horribly wrong with the manifest.

tests/set-working-directory_test.go
```diff
@@ -0,0 +1,48 @@
...
+	workingDir := setUpTest(t)
+	defer cleanUpTest(workingDir)
+
+	const wd = "/root"
+
+	_, _, _, err := runACBuild(workingDir, "set-working-dir", wd)
+	if err != nil {
+		t.Fatalf("%v\n", err)
+	}
+	man := schema.ImageManifest{
```
yifan-gu: can't use emptyManifestWithApp and set the wd then?

dgonyeo: That would require either undoing any changes I made to `emptyManifestWithApp` or finding a way to deep copy the struct. I'll look into it.

dgonyeo: actually, I'll just make `emptyManifestWithApp` a function that returns a new manifest, instead of just a var.

tests/set-event-handler_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"testing"
+
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema"
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema/types"
+)
+
+func manWithEH(eh types.EventHandler) schema.ImageManifest {
+	return schema.ImageManifest{
```
yifan-gu: ditto, use emptyManifestWithApp maybe?

tests/common.go
```diff
@@ -114,18 +150,18 @@ func runACBuild(workingDir string, args ...string) *acbuildError {
 	err = cmd.Wait()
 	if exitErr, ok := err.(*exec.ExitError); ok {
 		code := exitErr.Sys().(syscall.WaitStatus).ExitStatus()
-		return &acbuildError{exitErr, code, stdout, stderr}
+		return code, stdout, stderr, fmt.Errorf("non-zero exit code of %d: %v\nstdout:\n%s\nstderr:\n%s", code, exitErr, string(stdout), string(stderr))
```
yifan-gu: how about just override the stdout, stderr to a bytes.Buffer?
e.g. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/rkt/rkt.go#L200

tests/set-working-directory_test.go
yifan-gu: sgtm

tests/replace-manifest_test.go
yifan-gu: yea, but json.Marshal should invoke the same MarshalJSON function if it's been defined.
```If an encountered value implements the Marshaler interface and is not a nil pointer, Marshal calls its MarshalJSON method to produce JSON```
https://golang.org/pkg/encoding/json/#Marshal

dgonyeo: I was unaware of that, that's useful.

yifan-gu:
LGTM overall, just small nits.

PR #94, opened 2015-11-04
run: use the syscall package instead of execing mount/umount

dgonyeo:


jonboulle:
I feel like there's some unfortunate reason why this is a bad idea but can't think of why....

jonboulle:
Update README.md to remove mount reference

jonboulle:
Maybe false memory. LGTM after readme fix.

Issue #97, opened 2015-11-08
Not able to run container started with appc image built by acbuild


smakam:
Hi
I used the nginx example as below to create aci image in my CoreOS node.
acbuild begin
acbuild dependency add quay.io/fermayo/ubuntu
acbuild run -- apt-get update
acbuild run -- apt-get -y install nginx
acbuild set-exec /usr/sbin/nginx
acbuild set-name example.com/ubuntu-nginx
acbuild write ubuntu-nginx.aci
acbuild end

aci image got created successfully, but rkt container starts and exits. rkt version used is 0.9.0 and coreos version is 845.0.0. Following are the logs:
core@core-01 ~ $ sudo rkt run --debug --insecure-skip-verify ubuntu-nginx.aci
rkt: using image from file /usr/share/rkt/stage1-coreos.aci
rkt: using image from file /home/core/ubuntu-nginx.aci
rkt: searching for app image quay.io/fermayo/ubuntu
rkt: remote fetching from url https://quay.io/c1/aci/quay.io/fermayo/ubuntu/latest/aci/linux/amd64/
rkt: fetching image from https://quay.io/c1/aci/quay.io/fermayo/ubuntu/latest/aci/linux/amd64/
Downloading ACI: [=============================================] 72.4 MB/72.4 MB
2015/11/08 05:25:26 Preparing stage1
2015/11/08 05:25:27 Writing image manifest
2015/11/08 05:25:27 Loading image sha512-973e0f0a1c8252b1e92c882c902e4a00a0c75284678c39334f7c8a14d1a2740c
2015/11/08 05:25:33 Writing image manifest
2015/11/08 05:25:33 Writing pod manifest
2015/11/08 05:25:33 Setting up stage1
2015/11/08 05:25:33 Wrote filesystem to /var/lib/rkt/pods/run/5005b059-212e-4f68-82f3-ba4a7d5529c6
2015/11/08 05:25:33 Pivoting to filesystem /var/lib/rkt/pods/run/5005b059-212e-4f68-82f3-ba4a7d5529c6
2015/11/08 05:25:33 Execing /init
2015/11/08 05:25:33 Loading networks from /etc/rkt/net.d
2015/11/08 05:25:33 Loading network default with type ptp
Spawning container rkt-5005b059-212e-4f68-82f3-ba4a7d5529c6 on /var/lib/rkt/pods/run/5005b059-212e-4f68-82f3-ba4a7d5529c6/stage1/rootfs.
[...]

Issue #99, opened 2015-11-08
Nginx example produces ACI with empty rootfs


ericchiang:
Running the example in the README created an ACI with nothing in the rootfs. Not sure if this is a bug or if I'm missing something.

Here's my build file 

```
$ cat build.sh 
#!/bin/bash

acbuild begin
acbuild dependency add quay.io/fermayo/ubuntu
acbuild run -- apt-get update
acbuild run -- apt-get -y install nginx
acbuild set-exec /usr/sbin/nginx
acbuild set-name example.com/ubuntu-nginx
acbuild write ubuntu-nginx.aci
acbuild end
```

After building the ACI file and attempting to run it I got the following error message.

```
$ sudo rkt run --insecure-skip-verify ubuntu-nginx.aci 
rkt: using image from file /usr/bin/stage1-coreos.aci
rkt: using image from file /home/eric/p/acbuild/ubuntu-nginx.aci
rkt: using image from local store for image name quay.io/fermayo/ubuntu
[ 8331.672240] nginx[4]: Error: Unable to open "/usr/sbin/nginx": No such file or directory
```

Further inspection of the ACI file shows an empty rootfs directory

```
$ tar -tvf ubuntu-nginx.aci 
drwxr-xr-x 1000/1000         0 2015-11-08 12:54 rootfs
-rw-r--r-- root/root       271 2015-11-08 12:54 manifest
```

The manifest file looks pretty reasonable

```
$ tar -xf ubuntu-nginx.aci manifest -O | jq .
{
  "acKind": "ImageManifest",
  "acVersion": "0.7.1+git",
  "name": "example.com/ubuntu-nginx",
  "labels": [
    {
      "name": "arch",
      "value": "amd64"
    },
    {
      "name": "os",
      "value": "linux"
    }
  ],
  "app": {
    "exec": [
      "/usr/sbin/nginx"
    ],
    "user": "0",
    "group": "0"
  },
  "dependencies": [
    {
      "imageName": "quay.io/fermayo/ubuntu"
    }
  ]
}
```

Running Fedora 23. Here's some additional version data

```
$ sudo acbuild version
acbuild version v0.1.1-38-g0e884ec
appc version 0.7.1+git
$ sudo rkt version
rkt version 0.10.0
appc version 0.7.1
$ sudo systemd-nspawn --version
systemd 222
[...]

PR #100, opened 2015-11-09
docs: refactored README into Documentation directory

dgonyeo:
README was tweaked and greatly shortened. Documentation directory was
added, which includes a getting started guide, a description of context
free mode, and a suggested template for acbuild scripts. Each subcommand
got its own markdown file detailing its features and quirks.

Fixes https://github.com/appc/acbuild/issues/97 https://github.com/appc/acbuild/issues/99

### DIFF
--- a/Documentation/bash-scripts.md
+++ b/Documentation/bash-scripts.md
@@ -1,0 +1,39 @@
+# Bash scripts
+
+Bash scripting provides a common, simple, and well-known mechanism to drive
+`acbuild`. A script template, below, is used for the examples in this
+repository. It makes working with acbuild a little nicer by exiting the script
+when it hits an error, and calling `acbuild end` whenever the script exits.
+
+Since the script leverages `bash` features, we open it by specifying execution
+in the bash shell, rather than just inheriting the user's `SHELL` environment.
+
+Further, we set the `-e` option to bash to ensure that the entire script exits
+on the failure of any command. This gives us some atomicity, ensuring that
+either a complete and valid ACI is constructed, or none at all
+
+The `begin` line starts the build. If the build is to be started from an
+existing ACI, this line will be different.
+
+The rest of the script is concerned with cleanup and error handling. The
+`acbuildend` function is interesting, as it serves as a simple "catch" mechanism
+for the script, called on a trap triggered by either reaching the script's end,
+or by a non-zero return value from any command in the script. `acibuildend`
+stores the last command's exit code, then calls `acbuild --debug end` to
+terminate the build, passing the exit code, if any, through the script exit, to
+help with troubleshooting.
+
+```bash
+#!/usr/bin/env bash
+set -e
+
+acbuildend () {
+    export EXIT=$?;
+    acbuild --debug end && exit $EXIT;
+}
+
+acbuild --debug begin
+trap acbuildend EXIT
+
+# User entered acbuild commands go here
+```
--- a/Documentation/getting-started.md
+++ b/Documentation/getting-started.md
@@ -1,0 +1,143 @@
+# Getting Started with acbuild
+
+acbuild aims to support a workflow very similar to `Dockerfile`s, but with more
+flexibility and adherence to the Unix tools philosophy, and native image output
+in the modern ACI format. Generally, the workflow consists of a shell script
+(analogous to the `Dockerfile`) driving acbuild to construct an ACI, either
+from scratch or with another image as a base. This ACI is then stored in an
+_image registry_, and fetched by users to execute the applications stored
+within.
+
+The following guide will walk you through building and running a simple ACI
+with acbuild. The ACI will contain nginx, and will serve static files over HTTP.
+
+Note that some of the commands shown here must be run as root.
+
+## Making the ACI
+
+### Starting the build
+
+All acbuilds happen within a _context_. This build context is explicitly
+created at `acbuild begin`, and deleted at build completion with `acbuild end`.
+By default, the context and build are stored in the current working directory.
+
+This means that if you change directories in the middle of a build, acbuild will
+forget everything about the current build until you go back to the directory
+the build was started in. If that's an issue, you can override this with the
+`--work-path` flag, which will use the provided directory to store and access
+build information instead of the current directory.
+
+The command to start the build is:
+
+```bash
+acbuild begin
+```
+
+When `begin` isn't passed any arguments the build starts with an empty ACI.
+acbuild can also use an existing ACI as the starting point for the build.  More
+information on this feature is in the documentation for `begin`.
+
+### Naming the ACI
+
+All ACIs must be named. When a build is started with an empty ACI acbuild
+actually gives your ACI a placeholder name, and it won't let you write out an
+ACI until you change it away from that. If you're going to host this ACI
+somewhere the name should match the name used during meta discovery to find the
+download URL. For this example we'll just be running it off of the local
+filesystem, so name it whatever you want.
+
+```bash
+acbuild set-name example.com/nginx
+```
+
+For more information on meta discovery, check out the [appc
+spec](https://github.com/appc/spec/blob/master/spec/discovery.md#meta-discovery).
+
+### Adding a dependency
+
+Having an empty ACI isn't all that useful, so let's add a dependency.
+
+```bash
+acbuild dependency add quay.io/coreos/alpine-sh
+```
+
+Now when this ACI gets run the image at `quay.io/coreos/alpine-sh` will be
+fetched, and used as a base for our image. This means that our ACI only needs
+to contain the files that we add or modify on top of alpine.
+
+### Installing nginx
+
+Now that we've got a base image with fancy things like a shell and a package
+manager, let's install nginx.
+
+```bash
+acbuild run -- apk update
+acbuild run -- apk add nginx
+```
+
+The first command updates alpine's package manager, and the second command
+fetches and installs nginx. The `--` in each command stops acbuild from trying
+to parse flags that occur after the `--`. They're not necessary in this
+instance, but they would be if one of our commands had any arguments that began
+with a `-`.
+
+### Adding a mount point
+
+Since the goal of this is to host a static website, we're going to add a mount
+point to the default location that nginx serves files out of. This can be used
+when the ACI is run to make a directory on the host available inside of the
+container.
+
+```bash
+acbuild mount add html /usr/share/nginx/html
+```
+
+### Setting the exec statement
+
+The last thing we need to set in our ACI is the executable we want it to run.
+Without this command, the ACI couldn't be used without specifying a path to an
+executable inside of it at runtime.
+
+```bash
+acbuild set-exec -- /usr/sbin/nginx -g "daemon off;"
+```
+
+The `--` argument serves the same purpose here as it did when we were using the
+`run` command, except it actually matters this time. Without it, acbuild would
+think that the `-g` argument was a flag for it, instead of something to pass to
+`/usr/sbin/nginx` when the ACI is run.
+
+### Writing out the ACI
+
+With that, our ACI is complete. Before ending the build however, we need to
+write out the ACI to a file.
+
+```bash
+acbuild write nginx.aci
+```
+
+### Ending the build
+
+This command will delete our current build context. Had we not written out an
+ACI, everything we've done up to this point would be lost.
+
+```bash
+acbuild end
+```
+
+## Using the ACI
+
+If you have a directory with some files for nginx to serve...
+
+```bash
+mkdir test
+echo "Hello, world!" > test/index.html
+```
+
+... using the ACI is something as simple as the following command:
+
+```bash
+rkt run --insecure-skip-verify ./nginx.aci --volume html,kind=host,source=/path/to/test --net=host
+```
+
+Now point your browser at [`http://localhost`](http://localhost).
--- a/Documentation/subcommands/dependency.md
+++ b/Documentation/subcommands/dependency.md
@@ -1,0 +1,49 @@
+# acbuild dependency
+
+Dependencies are ACIs separate from the current ACI, that are placed down into
+the rootfs before the files from the current ACI.
+
+The ordering of an ACI's dependencies is significant. Let's say the current ACI
+has two dependencies, A and B, where A comes before B. If both A and B contain
+the file `/bin/sh`, and the current ACI does not have this file, then the
+`/bin/sh` from B is what will appear in the container when it is run.
+
+At the time of writing acbuild simply puts the dependency in the manifest in
+the order they were added. If the ordering of an ACI's dependencies is
+incorrect, the user must remove each dependency manually and then add them in
+the correct order.
+
+## Subcommands
+
+* `acbuild dependency add IMAGE_NAME`
+
+  Updates the ACI to contain a dependency with the given name.
+
+* `acbuild dependency remove IMAGE_NAME`
+
+  Removes the dependency with the given image name from the ACI.
+
+## Flags
+
+The `add` command also has the following optional flags:
+
+- `--image-id`: sets the content hash of the dependency being added. When this
+  ACI is run, the retrieved dependency must match this hash.
+
+- `--label`: adds a label to the dependency being added. This is used when
+  determining which image to fetch for the dependency.
+
+- `--size`: the size of the dependency being added, in bytes. When this ACI is
+  run, the retrieved dependency must have this size.
+
+## Examples
+
+```bash
+acbuild dependency add example.com/alpine
+
+acbuild dependency add example.com/ubuntu --image-id sha512-...
+
+acbuild dependency add example.com/nodejs --label version=4.0.0 --label arch=noarch
+
+acbuild dependency remove example.com/centos
+```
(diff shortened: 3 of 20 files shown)


dgonyeo:
@alban I added documentation for a common workflow, but it's not very good right now. If you have any ideas about what should go in there I'd love to hear them.

jonboulle:
ping @joshix for help reviewing..

README.md
```diff
@@ -6,14 +6,13 @@ acbuild is a command line utility to build and modify App Container Images
 
 ## Rationale
 
-Dockerfiles are powerful and feature useful concepts such as build layers, 
-controlled build environment. At the same time, they lack flexibility 
+Dockerfiles are powerful, and feature useful concepts such as build layers and
```
joshix: Is it actually necessary to mention `Dockerfile`s at all? 

If so, is what you want to say actually "We wanted a flexible, extensible tool for manipulating ACIs, well-integrated with the Unix toolkit, and as powerful and simple as the `Dockerfile` mechanism."

README.md
```diff
@@ -6,14 +6,13 @@ acbuild is a command line utility to build and modify App Container Images
...
+a controlled build environment. At the same time, they lack flexibility 
 (impossible to extend, re-use environment variables) and don't play nicely 
 with the appc spec and Linux toolchain (e.g. shell, makefiles)
 
-`acbuild` is a command-line tool that natively supports ACI builds and integrates 
-well with shell, makefiles and
-other Unix tools.
+`acbuild` is a command-line tool that natively supports ACI builds and
+integrates well with shell, makefiles and other Unix tools.
```
joshix: ...the shell, `Makefile`s, and other Unix tools.

README.md
```diff
@@ -28,9 +27,15 @@ which it's run:
 - `modprobe`
 - `gpg`
 
+### Prebuilt Binaries
+
+The easiest way to get `acbuild` is to download one of the releases from
```
joshix: The easiest way to get `acbuild` is to download one of the [releases](https://github.com/appc/acbuild/releases) from Github.

README.md
dgonyeo: I'd be in favor of not mentioning them at all. That text is leftover from the original README.

Perhaps:
We wanted a powerful tool for manipulating ACIs, that made it easy to iteratively build images both from scratch and on top of existing images, and would play nicely with popular Linux tools like the shell and `Makefile`s.

Might need rephrasing and/or more points tossed in there though.

Documentation/bash-scripts.md
```diff
@@ -0,0 +1,52 @@
+# Bash scripts
+
+A common way to use acbuild is via a bash script. There's a common template
```
joshix: Bash scripting provides a common, simple, and well-known mechanism to drive `acbuild`s. A script template, below, is used for the examples here.

Documentation/bash-scripts.md
```diff
@@ -0,0 +1,52 @@
...
+that the examples for acbuild follow, and it makes cleanup a little easier. The
```
joshix: (Either explain how the template "makes cleanup easier", or cut that clause entirely.)

Documentation/bash-scripts.md
```diff
@@ -0,0 +1,52 @@
...
+template is at the end of this file, and here we explain the reasoning behind
```
joshix: Cut lines 5 and 6

Documentation/bash-scripts.md
```diff
@@ -0,0 +1,52 @@
...
+the parts of it.
+
+The 1st line, `#!/usr/bin/env bash`, is simply good form to make sure the
```
joshix: Since the script leverages `bash` features, we open it by specifying execution in the `bash` shell, rather than just inheriting the user's `SHELL` enviroment.

Further, we set the `-e` option to `bash` to ensure that the entire script exits on the failure of any command. This gives us some atomicity, ensuring that either a complete and valid ACI is constructed, or none at all.

Documentation/bash-scripts.md
```diff
@@ -0,0 +1,52 @@
...
+script is executed with bash as opposed to the current shell.
+
+The 2nd line, `set -e` makes the script exit in the event of an error. If some
+command doesn't exit correctly, due to something like a typo in a command, or
+the command in `acbuild run` encountering an error, it is better to end the
+script, instead of continuing on and producing an ACI with unintended values or
+contents.
+
+The 4th through the 7th lines check if the script is being run as the root
```
joshix: If we're `-e` bailing on any nonzero return, why can't we cut this paragraph and these lines from the script? If I'm not root, script will fail and bail if it turns out I need to be root for some step, right? In other words, can we just leave the user handling to the environment. Make any sense?

dgonyeo: Yeah, that would be fine. I thought it would be a little nicer to get the error up front, as opposed to seeing a handful of commands succeed before running into a "you need to be root" error, but it might not be worth the added complexity.

README.md
```diff
@@ -28,9 +27,15 @@ which it's run:
...
 
+### Prebuilt Binaries
+
+The easiest way to get `acbuild` is to download one of the
+[releases](https://github.com/appc/acbuild/releases) from GitHub.
+
 ### Build from source
 
-Currently the only way to install `acbuild` is to build from source. 
+The other way to get `acbuild` is to build it from source.
```
joshix: Might mention the platform supported/required, which far as I can tell is only Linux amd64 (right? --at any rate I can't build it on darwin)

Documentation/bash-scripts.md
```diff
@@ -0,0 +1,52 @@
...
+#!/usr/bin/env bash
+set -e
+
+if [ "$EUID" -ne 0 ]; then
+    echo "This script uses functionality which requires root privileges"
+    exit 1
+fi
+
+acbuildend () {
+    export EXT=$?;
```
joshix: Unless it conflicts with something else in the env, I'd just name this var `EXIT`

Documentation/bash-scripts.md
```diff
@@ -0,0 +1,52 @@
...
+command doesn't exit correctly, due to something like a typo in a command, or
+the command in `acbuild run` encountering an error, it is better to end the
+script, instead of continuing on and producing an ACI with unintended values or
+contents.
+
+The 4th through the 7th lines check if the script is being run as the root
+user. If it is not, the script prints an error and exits. This check may not be
+necessary, depending on the actions performed in the script.
+
+The 9th through 12th lines define a function that performs cleanup after the
```
joshix: The rest of the script is concerned with cleanup and error handling. The `acbuildend` function is interesting, as it serves as a simple "catch" mechanism for the script, called on a `trap` triggered by either reaching the script's end, or  by a non-zero return value from any command in the script. `Acibuildend` stores the last command's exit code, then calls `acbuild end` to terminate the build, passing the exit code, if any, through the script exit, to help with troubleshooting.

Documentation/bash-scripts.md
```diff
@@ -0,0 +1,52 @@
...
+
+The 14th line starts the build. If the build is to be started from an existing
+ACI, this line will be different.
+
+The 15th line sets up a trap to call the `acbuildend` function when the script
+exits. This will happen either if the end of the script is reached, or if the
+script is ending due to a command not exiting cleanly. The function it performs
+will take note of the exit code of the last command that was run, and then call
+`acbuild end`. If `acbuild end` exits cleanly, then it exits with the exit code
+it took note of.
```
joshix: (Suggestion is to repl lines 21-35 with the text in my previous comment, above.)

Documentation/getting-started.md
```diff
@@ -0,0 +1,134 @@
+# Getting Started with acbuild
+
+The following guide will walk you through building and running a simple ACI
+with acbuild. The ACI will contain nginx, and be used to host static files.
```
joshix: s/host static files/serve static files over HTTP/

Documentation/getting-started.md
```diff
@@ -0,0 +1,134 @@
+# Getting Started with acbuild
+
```
joshix: acbuild aims to support a workflow very similar to `Dockerfile`s, but with more flexibility and adherence to the Unix tools philosophy, and native image output in the modern ACI format. Generally, the workflow consists of a shell script (analogous to the `Dockerfile`) driving acbuild to construct an ACI, either from scatch or with another image as a base. This ACI is then stored in an *image registry*, and fetched by users to execute the applications stored within.

Documentation/common-workflow.md
```diff
@@ -0,0 +1,6 @@
+# Common Workflow
```
joshix: Suggest rm Documentation/common-workflow.md after inserting this same info in Documentation/getting-started as per my comment on that file, graf 2, below.

Documentation/getting-started.md
```diff
@@ -0,0 +1,134 @@
...
+The following guide will walk you through building and running a simple ACI
+with acbuild. The ACI will contain nginx, and be used to host static files.
+
+Note that some of the commands shown here must be run as root.
+
+## Making the ACI
+
+### Starting the build
+
+All acbuild builds have a context, that needs to be explicitly created when the
```
joshix: All acbuilds happen within a *context*. This build context is explicitly created at `acbuild begin`, and deleted at build completion with `acbuild end`. By default, the context and build are stored in the current working directory.

Documentation/getting-started.md
```diff
@@ -0,0 +1,134 @@
...
+build is started and deleted when the build is finished. It defaults to being
+stored in your current working directory, so if you change directories in the
```
joshix: Newline after period, before "so if you change directories..."

Then rewrite from "so if you..." for clarity (JW:TODO: Write suggested replacement)

dgonyeo:
Made the changes you suggested. I'm going to add changes resulting from this review into new commits, so you can see what I end up changing between passes. I'll squash them together when I get the "LGTM".

joshix:
Thanks for doing that. I was going to ask that you hold your squash for the end but didn't want to seem too nitpicky.

dgonyeo:
I may have squashed a tiny amount of changes I made yesterday before realizing I should do this, but it probably wasn't much.

Documentation/getting-started.md
```diff
@@ -0,0 +1,143 @@
...
+
+All ACIs must be named. When a build is started with an empty ACI acbuild
+actually gives your ACI a placeholder name, and it won't let you write out an
+ACI until you change it away from that. If you're going to host this ACI
+somewhere the name should match the name used during meta discovery to find the
+download URL. For this example we'll just be running it off of the local
+filesystem, so name it whatever you want.
+
+```bash
+acbuild set-name example.con/nginx
```
alban: example.co**m**

Documentation/getting-started.md
```diff
@@ -0,0 +1,143 @@
...
+
+```bash
+mkdir test
+echo "Hello, world!" > test/index.html
+```
+
+... using the ACI is something as simple as the following command:
+
+```bash
+rkt run --insecure-skip-verify ./nginx.aci --volume html,kind=host,source=/path/to/test --net=host
```
alban: This is good for now, but it will need to be changed after https://github.com/coreos/rkt/issues/912.

/cc @blixtra

Documentation/getting-started.md
```diff
@@ -0,0 +1,143 @@
...
+think that the `-g` argument was a flag for it, instead of something to pass to
+`/usr/sbin/nginx` when the ACI is run.
+
+### Writing out the ACI
+
+With that, our ACI is complete. Before ending the build however, we need to
+write out the ACI to a file.
+
+```bash
+acbuild write nginx.aci
```
alban:
```
$ acbuild write nginx.aci
write: open .acbuild/currentaci/rootfs/etc/shadow: permission denied
```

I wonder if the error is good enough, or if `acbuild` should say you must be root.

Documentation/getting-started.md
```diff
@@ -0,0 +1,143 @@
...
+in the modern ACI format. Generally, the workflow consists of a shell script
+(analogous to the `Dockerfile`) driving acbuild to construct an ACI, either
+from scratch or with another image as a base. This ACI is then stored in an
+_image registry_, and fetched by users to execute the applications stored
+within.
+
+The following guide will walk you through building and running a simple ACI
+with acbuild. The ACI will contain nginx, and will serve static files over HTTP.
+
+Note that some of the commands shown here must be run as root.
```
alban: The documentation could also start the commands with `$` or `#` to indicate if `root` is necessary.

Documentation/getting-started.md
```diff
@@ -0,0 +1,143 @@
...
+echo "Hello, world!" > test/index.html
+```
+
+... using the ACI is something as simple as the following command:
+
+```bash
+rkt run --insecure-skip-verify ./nginx.aci --volume html,kind=host,source=/path/to/test --net=host
+```
+
+Now point your browser at [`http://localhost:8888`](http://localhost:8888).
```
alban: On my computer, nginx started on port 80 and not 8888. So it should just be `http://localhost/` or `http://localhost:80/`.

Also, I get a "403 Forbidden" error, and strace on nginx says:
```
open("/usr/share/nginx/html/index.html", O_RDONLY|O_NONBLOCK) = -1 EACCES (Permission denied)
```

Indeed, the process nginx runs as user `100` (and not `root`) and it does not have `x` permission on `/`:
```
# ls -ld  /proc/`pidof 'nginx: worker process'`/root/
drw-r--r--. 1 root root 4096 Nov 14 16:24 /proc/970/root/
```

This comes from the alpine image, as reported on https://github.com/coreos/rkt/issues/1703#issuecomment-152848856: 
```
$ wget -O - https://quay.io/c1/aci/quay.io/coreos/alpine-sh/latest/aci/linux/amd64/ | tar tvz|head
drw-r--r-- 0/0               0 1970-01-01 01:00 rootfs/
```

Documentation/subcommands/dependency.md
```diff
@@ -0,0 +1,49 @@
...
+
+The ordering of an ACI's dependencies is significant. Let's say the current ACI
+has two dependencies, A and B, where A comes before B. If both A and B contain
+the file `/bin/sh`, and the current ACI does not have this file, then the
+`/bin/sh` from B is what will appear in the container when it is run.
+
+At the time of writing acbuild simply puts the dependency in the manifest in
+the order they were added. If the ordering of an ACI's dependencies is
+incorrect, the user must remove each dependency manually and then add them in
+the correct order.
```
alban: Is there a way to list the current dependencies or to remove all dependencies if the user forgets the dependencies? The current way seems to be:

```
json_pp < .acbuild/currentaci/manifest
```

README.md
```diff
@@ -67,188 +78,28 @@ and put the following lines at the end of the file:
...
-
-When acbuild goes to write the ACI, meta discovery could be performed on the
-name that was set for the ACI to find an endpoint to push the image to, instead
-of saving the ACI on the local filesystem.
-
-### Alternate execution engine
-
-Support multiple execution engines, notably runc.
+Documentation about acbuild and many of its commands is available in the
+`Documentation` directory in this repository.
```
alban: can we have a link?

Documentation/getting-started.md
dgonyeo: I think that the error is clear enough, many things in Linux spit back `permission denied` when something needs to be run as root.

Documentation/getting-started.md
dgonyeo: The permissions issue on the alpine image _should_ be fixed as of https://github.com/coreos-inc/quay/pull/803. It's possible that the ACI is still cached from before the fix though.

Documentation/subcommands/dependency.md
dgonyeo: The dependencies can be viewed with `acbuild cat-manifest --pretty-print`. Perhaps commands for clearing and printing just the list would be in order, but then I'd want to consider adding similar commands for mounts/ports/labels/etc.

Documentation/getting-started.md
jonboulle: Eh, I really don't like this UX and I'd rather we went back to the pre-check.

jonboulle:
I'm inclined to land this and roll forward, any objections @joshix ?

Documentation/getting-started.md
dgonyeo: By "pre-check" do you mean checking before write is called? Because write will not need root depending on what's in the ACI. I suppose that write could walk the file tree checking if it can't access anything before actually doing the write, if you want a nicer error message in this situation.

dgonyeo: Or I could find what type of error `github.com/coreos/rkt/pkg/fileutil.CopyTree` returns when it gets permission denied, which would be less work and probably saner.

joshix:
Nope, @jonboulle @dgonyeo LGTM.

(Also dgonyeo I formally withdraw any initial objection to subcmd-per-file I brought up verbally last week)

Documentation/getting-started.md
dgonyeo: I just tweaked `lib/write.go`, and it now catches this error and emits something phrased as the following:
```
write: "/test": permission denied - call write as root
```

README.md
jonboulle: @dgonyeo hmm, seems this got lost before merge

dgonyeo: Whoops, I'll make a PR for that tomorrow.

PR #101, opened 2015-11-11
begin: require first char to be / or . when starting from local file

dgonyeo:
So as to provide better error messages when a user attempts to start a
build from a local file and this file doesn't exist, the user must
provide a path to the file starting with `/` or `.`.

Additionally, begin now properly cleans up whatever context it started
to make if an error is returned.

### DIFF
--- a/lib/begin.go
+++ b/lib/begin.go
@@ -17,6 +17,8 @@ package lib
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
+	"net/url"
 	"os"
 	"path"
 	"runtime"
@@ -64,12 +66,24 @@ func (a *ACBuild) Begin(start string, insecure bool) (err error) {
 		}
 	}()
 
+	defer func() {
+		// If there was an error while beginning, we don't want to produce an
+		// unexpected build context
+		if err != nil {
+			os.RemoveAll(a.ContextPath)
+		}
+	}()
+
 	if start != "" {
 		err = os.MkdirAll(a.CurrentACIPath, 0755)
 		if err != nil {
 			return err
 		}
-		return a.beginFromImage(start, insecure)
+		if start[0] == '.' || start[0] == '/' {
+			return a.beginFromLocalImage(start)
+		} else {
+			return a.beginFromRemoteImage(start, insecure)
+		}
 	}
 
 	err = os.MkdirAll(path.Join(a.CurrentACIPath, aci.RootfsDir), 0755)
@@ -131,25 +145,23 @@ func (a *ACBuild) Begin(start string, insecure bool) (err error) {
 	return nil
 }
 
-func (a *ACBuild) beginFromImage(start string, insecure bool) error {
-	// Check if we're starting with a file
+func (a *ACBuild) beginFromLocalImage(start string) error {
 	finfo, err := os.Stat(start)
 	if err == nil {
 		if finfo.IsDir() {
 			return fmt.Errorf("provided starting ACI is a directory: %s", start)
 		}
 		return util.ExtractImage(start, a.CurrentACIPath, nil)
-	} else if !os.IsNotExist(err) {
-		return err
 	}
+	return err
+}
 
+func (a *ACBuild) beginFromRemoteImage(start string, insecure bool) error {
 	// Check if we're starting with a docker image
 	if strings.HasPrefix(start, "docker://") {
 		// TODO use docker2aci
 		return fmt.Errorf("docker containers are currently unsupported")
 	}
-
-	// Perform meta discovery, download the ACI, and start with that.
 
 	app, err := discovery.NewAppFromString(start)
 	if err != nil {
@@ -181,6 +193,15 @@ func (a *ACBuild) beginFromImage(start string, insecure bool) error {
 
 	err = reg.Fetch(app.Name, labels, 0, false)
 	if err != nil {
+		if urlerr, ok := err.(*url.Error); ok {
+			if operr, ok := urlerr.Err.(*net.OpError); ok {
+				if dnserr, ok := operr.Err.(*net.DNSError); ok {
+					if dnserr.Err == "no such host" {
+						return fmt.Errorf("unknown host when fetching image, check your connection and local file paths must start with '/' or '.'")
+					}
+				}
+			}
+		}
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


dgonyeo:
If this gets merged before the docs PR, I'll update that to reflect the new semantics of begin. If the docs PR gets merged first, I'll rebase and add those changes here.

lib/begin.go
```diff
@@ -179,6 +189,9 @@ func (a *ACBuild) beginFromImage(start string, insecure bool) error {
 
 	err = reg.Fetch(app.Name, labels, 0, false)
 	if err != nil {
+		if strings.Contains(err.Error(), "no such host") {
```
jonboulle: :/, could we compare against net.DNSError or something instead?

dgonyeo: I guess checking the string isn't ideal. Just pushed a commit, I now do a type assertion on err to see if it's a `*url.Error`, and if it's ok return the helpful error.

lib/begin.go
```diff
@@ -179,6 +190,9 @@ func (a *ACBuild) beginFromImage(start string, insecure bool) error {
 
 	err = reg.Fetch(app.Name, labels, 0, false)
 	if err != nil {
+		if _, ok := err.(*url.Error); ok {
+			return fmt.Errorf("unknown host when fetching image, local file paths must start with '/' or '.'")
```
jonboulle: but are we sure it's an unknown host error? This feels like it might be catching too generously

dgonyeo: Ah yep. This also catches the error returned when network is missing. Any idea on how I could only catch when DNS lookup fails? Maybe I should just try to lookup the domain before handing it off to `appc/spec/discovery`, and print the special error message when that fails, but I don't know if it'll be any easier there to differentiate between the network's down and the host doesn't exist.

Maybe a special error message is unnecessary if the documentation / help page for the begin command describe the quirk here?

jonboulle: Cant you check against the dns error type like I first suggested?

dgonyeo: That didn't appear to be the error returned when I did something like `acbuild begin alpine-latest-linux-amd64.aci`.

I got `*url.Error` from a `fmt.Printf("%s\n", reflect.TypeOf(error).String())` just before the check.

jonboulle: Weird, what is the content of the error?

An alternative approach is to actually do a stat and see if there is a file they were trying to reference.. I don't know if I like that though.

dgonyeo:
```
        if urlerr, ok := err.(*url.Error); ok {                                     
            fmt.Printf("Op: %s\nURL: %s\nErr: %s\n", urlerr.Op, urlerr.URL, urlerr.Err.Error())
            return fmt.Errorf("unknown host when fetching image, local file paths must start with '/' or '.'")
        }
```
```
derek@energia ~/go/src/github.com/appc/acbuild> ./bin/acbuild begin lib/begin.go 
Op: Get
URL: https://lib?ac-discovery=1
Err: dial tcp: lookup lib: no such host
begin: unknown host when fetching image, local file paths must start with '/' or '.'
```

dgonyeo: I followed the rabbit hole down as far as I could, and I think the check I have for it is as best I can get. Notably the exact same error gets returned for both when there is no internet connection and when there is a connection and the DNS lookup failed. I've modified the returned error to reflect this.

dgonyeo:
I just updated the docs to reflect this change, so please also make sure the phrasing/spelling there is sane.

lib/begin.go
```diff
@@ -181,6 +193,15 @@ func (a *ACBuild) beginFromImage(start string, insecure bool) error {
 
 	err = reg.Fetch(app.Name, labels, 0, false)
 	if err != nil {
+		if urlerr, ok := err.(*url.Error); ok {
+			if operr, ok := urlerr.Err.(*net.OpError); ok {
+				if dnserr, ok := operr.Err.(*net.DNSError); ok {
+					if dnserr.Err == "no such host" {
```
jonboulle: I regret sending you down this path..

dgonyeo: It even still boils down to a string comparison. I was really hoping for a `net.ErrNoSuchHost`.

yifan-gu: @jonboulle @dgonyeo  what's the reason for this ?

yifan-gu:
I guess I missed the reason why this is better than the old behavior (which check if the image exists on local first, if not treat it as a remote image. Also this is consistent with `rkt fetch`).
And this is preventing me from providing `~/github.com/foo/bar/baz.aci`, though not a big deal :)

lib/begin.go
dgonyeo: I'm modifying begin such that when a local file is specified as a starting point, it must begin with `/` or `.`. This is so that I can return a nicer error message when the local file isn't found. This introduces an issue of detecting when the user wants to use a local file but doesn't start the argument with the correct character. The way acbuild handles this is attempting to do meta discovery on the string and then failing when the DNS lookup fails. This is attempting to determine when that error happens. @jonboulle didn't like me doing `strings.Contains(err.Error(), "no such host")`, so this was the result.

dgonyeo:
Oh yeah, `~` should definitely be an allowed starting character.

lib/begin.go
yifan-gu: @dgonyeo I see. Thank you, maybe we should do the same in `rkt fetch`...

jonboulle: Reminder about https://github.com/coreos/rkt/issues/715

yifan-gu: @jonboulle Thanks, you memorize every issue! 
So also adding checking for `..` here?

ecnahc515:
Shouldn't `~` get interpreted by the shell before it gets to rkt?

lib/begin.go
dgonyeo: .. Starts with . so it's covered by the current rules.

dgonyeo:
Either way the docs should state that it's allowed.

dgonyeo:
Just tweaked the docs to also mention `~`. Once this gets merged I'd like to release v0.2.0. Anything else you'd like to see done to this @jonboulle?

dgonyeo:
Oh, and to answer Yifan's "I missed the reason why this is better than the old behavior", this is a UX thing. I can't provide a good error message when you typo a local file name, since if the file isn't local it'll do meta discovery, and then you'd get "no such host" when you typo a file name.

jonboulle:
:sheep:

PR #102, opened 2015-11-11
write: when an error is encountered delete half-written ACI

dgonyeo:


lib/write.go
```diff
@@ -72,6 +72,14 @@ func (a *ACBuild) Write(output string, overwrite, sign bool, gpgflags []string)
 	}
 	defer ofile.Close()
 
+	defer func() {
+		// When write is done, if an error is encountered remove the partial
+		// ACI that had been written.
+		if err != nil {
+			os.RemoveAll(output)
```
jonboulle: Note this is redundant with 90/98 (and unclear why inconsistent, RemoveAll here and Remove there)

dgonyeo: So it is! I removed the removes at 90/98, added a removal for `output + "asc."` here, and switched to Remove here. In this context, Remove and RemoveAll are equivalent afaik.

jonboulle:
Sure, lgtm

PR #103, opened 2015-11-12
copy: allow for copying from multiple sources at once

dgonyeo:
This commit allows for multiple sources for a copy to be specified, so
`acbuild copy` behaves a little more like `cp`. If there are multiple
sources, the destination must be a directory. If there are multiple
sources and the destination directory doesn't exist, the directory is
created.

This allows the * expansion from the shell, which makes copying all the
contents of a directory to a location in the ACI much easier.

For example: `acbuild copy html/* /usr/share/nginx/html/`


acbuild/copy.go
```diff
@@ -20,9 +20,9 @@ import (
 
 var (
 	cmdCopy = &cobra.Command{
-		Use:     "copy PATH_ON_HOST PATH_IN_ACI",
+		Use:     "copy PATHS_ON_HOST PATH_IN_ACI",
```
jonboulle: PATH_ON_HOST...

acbuild/copy.go
```diff
@@ -20,9 +20,9 @@ import (
...
 		Short:   "Copy a file or directory into an ACI",
-		Example: "acbuild copy nginx.conf /etc/nginx/nginx.conf",
+		Example: "acbuild copy stuff/* nginx.conf /etc/nginx/",
```
jonboulle: an example for single file would be cool too

jonboulle:
test? :D

acbuild/copy.go
dgonyeo: Cobra only lets me set a single example. I'll put examples for both in the documentation, but which would you prefer goes here?

dgonyeo:
fsdiffer would actually be incredibly useful for writing tests for copy. I'm going to make a PR to add that, and once it gets merged I'll add some tests here.

acbuild/copy.go
jonboulle: Hmm. If you do a multi-line string here does it all go to hell?

acbuild/copy.go
```diff
@@ -20,9 +20,9 @@ import (
 
 var (
 	cmdCopy = &cobra.Command{
-		Use:     "copy PATH_ON_HOST PATH_IN_ACI",
+		Use:     "copy PATH_ON_HOST... PATH_IN_ACI",
```
jonboulle: I guess really we have two modes now like `cp`:
```
$ cp --help
Usage: cp [OPTION]... [-T] SOURCE DEST
  or:  cp [OPTION]... SOURCE... DIRECTORY
...
``` 
so really it should be more like `copy PATH_ON_HOST PATH_IN_ACI` and `copy PATH_ON_HOST... DIRECTORY_IN_ACI` - don't suppose we can have multilines here too?

dgonyeo: Maybe with some serious tweaking of the help page templates.

In their current state we get:

```bash
USAGE:
	acbuild copy PATH_ON_HOST PATH_IN_ACI
copy PATH_ON_HOST... DIRECTORY_IN_ACI
```

dgonyeo:
Test! :D

lib/copy.go
```diff
@@ -37,13 +38,42 @@ func (a *ACBuild) Copy(from, to string) (err error) {
...
 			return err
 		}
+	case err != nil:
+		return err
+	case !targetInfo.IsDir():
+		return fmt.Errorf("target %q is not a directory", to)
 	}
 
-	return fileutil.CopyTree(from, target, uid.NewBlankUidRange())
+	targets := make([]string, len(froms))
```
jonboulle: Why do this in two loops rather than one? targets seems superfluous.

tests/copy_test.go
```diff
@@ -0,0 +1,278 @@
...
+	defer cleanUpTest(workingDir)
+
+	sourceDir, err := ioutil.TempDir("", "acbuild-test-copy")
+	if err != nil {
+		panic(err)
+	}
+	defer os.RemoveAll(sourceDir)
+
+	time1 := time.Now()
+	//time2 := time.Now().Add(1 * time.Microsecond)
```
jonboulle: ?

tests/copy_test.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	froms := []string{"dir01", "dir02", "dir03"}
+	for i := 0; i < len(froms); i++ {
+		froms[i] = path.Join(sourceDir, froms[i])
+	}
+
+	// golang--
```
jonboulle: can you explain this? I don't understand the issue

tests/copy_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+	time1 := time.Now()
+	//time2 := time.Now().Add(1 * time.Microsecond)
+	files := []*buildFileInfo{
+		&buildFileInfo{path: "file01", typeflag: tar.TypeReg, mode: 0644, atime: time1, mtime: time1, contents: "hello"},
+		&buildFileInfo{path: "dir01", typeflag: tar.TypeDir, mode: 0755, atime: time1, mtime: time1},
+		&buildFileInfo{path: "dir01/file01", typeflag: tar.TypeReg, mode: 0644, atime: time1, mtime: time1, contents: "hello"},
+	}
+
+	err = buildFS(sourceDir, files)
```
jonboulle: Seems like you probably just want this to be mustBuildFS()

tests/copy_test.go
```diff
@@ -0,0 +1,278 @@
...
+	defer cleanUpTest(workingDir)
+
+	sourceDir, err := ioutil.TempDir("", "acbuild-test-copy")
+	if err != nil {
+		panic(err)
+	}
+	defer os.RemoveAll(sourceDir)
+
+	time1 := time.Now()
+	//time2 := time.Now().Add(1 * time.Microsecond)
```
jonboulle: ?

tests/copy_test.go
```diff
@@ -0,0 +1,278 @@
...
+	defer os.RemoveAll(sourceDir)
+
+	err = os.MkdirAll(path.Join(workingDir, ".acbuild", "currentaci", aci.RootfsDir, dest), 0755)
+	if err != nil {
+		panic(err)
+	}
+
+	err1 := runACBuild(workingDir, "copy", sourceDir, dest)
+	if err1 == nil {
+		t.Fatalf("got no error when destination directory exists, was expecting one")
```
jonboulle: I'm confused by this one. Aren't we providing `cp -r`-like semantics?

jonboulle: Also, if possible this kind of test should check for the actual error type to avoid the risk of false positives

util/fsdiffer/beforeafter_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"testing"
+	"time"
+)
+
+// Do not test (at the moment) symlinks as every os needs a special syscall to set symlink times
+func TestBeforeAfterFSDiffer(t *testing.T) {
+	time1 := time.Now()
+	time2 := time.Now().Add(1 * time.Microsecond)
+	sourceFiles := []*buildFileInfo{
+		&buildFileInfo{path: "file01", typeflag: tar.TypeReg, mode: 0644, atime: time1, mtime: time1, contents: "hello"},
```
jonboulle: How about some helpers to clean these up:
`func mkBuildFileInfoDir(path string, time time.Time) *buildFileInfo`
`func mkBuildFileInfoFile(path string, time time.Time, contents string) *buildFileInfo`

util/fsdiffer/beforeafter.go
```diff
@@ -0,0 +1,63 @@
+package fsdiffer
+
+import (
+	"fmt"
+	"path/filepath"
+)
+
+// The BeforeAfterFSDiffer is used to generate changes in a given directory
```
jonboulle: s/The//
I don't particularly like the name BeforeAfterFSDiffer, but am struggling to think of an alternative right now..

util/fsdiffer/beforeafter.go
```diff
@@ -0,0 +1,63 @@
...
+
+// NewBeforeAfterFSDiffer creates a new BeforeAfterFSDiffer that will report
+// changes on the given directory.
+func NewBeforeAfterFSDiffer(dir string) *BeforeAfterFSDiffer {
+	return &BeforeAfterFSDiffer{dir: dir}
+}
+
+// Start will walk the provided directory, and record its state for use as the
+// before part of the changes returned by Diff.
+func (ba *BeforeAfterFSDiffer) Start() error {
```
jonboulle: I'm struggling to think of a use case for initialising a BeforeAfterFSDiffer and not immediately calling Start() - what about just collapsing this into the constructor?

acbuild/copy.go
jonboulle: I'm a little torn here; it would be great to be explicit about these different operation modes. Any other ideas?

dgonyeo: Could split the different behaviors out into different commands, but I'd want to be careful about naming to make sure it's clear why there are two separate commands. I'm still not sure why Dockerfiles have both Add and Copy.

util/fsdiffer/beforeafter.go
dgonyeo: Since it produced changes between two different points in time:
HistoricalFSDiffer
TemporalFSDiffer

jonboulle: Maybe Temporal

tests/copy_test.go
dgonyeo: I'm just annoyed that I can't do:
```go
err1 := runACBuild(workingDir, "copy", append(froms, dest)...)
```
or even:
```go
err1 := runACBuild(workingDir, "copy", froms..., dest)
```

But both of those result in: `too many arguments in call to runACBuild`.

To use the `...`, there can only be one argument for the entire `...` parameter in the function type, so the only way to pass in a string, a slice of strings, and another string as the list of arguments to acbuild here is to append the final string onto the slice of strings, and then to append that new slice onto the end of a singleton slice holding the first string. It's messy, and there's no reason that Go couldn't do this for me.

util/fsdiffer/beforeafter.go
dgonyeo: I made the change in https://github.com/appc/acbuild/pull/104. Once that gets merged I'll rebase this on master and rename the calls in here.

util/fsdiffer/beforeafter.go
dgonyeo: Done (also in https://github.com/appc/acbuild/pull/104)

dgonyeo:
This should be ready for another pass.

acbuild/copy.go
```diff
@@ -36,8 +38,8 @@ func runCopy(cmd *cobra.Command, args []string) (exit int) {
 		cmd.Usage()
 		return 1
 	}
-	if len(args) != 2 {
-		stderr("copy: incorrect number of arguments")
+	if len(args) < 2 || (explicitTarget && len(args) != 2) {
+		stderr("copy: incorrect number of arguments: %v", args)
```
jonboulle: why spit the arguments back out to them? isn't that obvious?

lib/copy.go
```diff
@@ -25,7 +26,7 @@ import (
 
 // Copy will copy the directory/file at from to the path to inside the untarred
```
jonboulle: needs update

lib/copy.go
```diff
@@ -37,13 +38,41 @@ func (a *ACBuild) Copy(from, to string) (err error) {
...
+		}
+		dir, _ := path.Split(target)
+		if dir != "" {
+			err := os.MkdirAll(dir, 0755)
+			if err != nil {
+				return err
+			}
+		}
+
+		return fileutil.CopyTree(froms[0], target, uid.NewBlankUidRange())
```
jonboulle: Should we use CopyRegularFile here? is that all we support in this mode?

tests/copy_test.go
jonboulle: oh, I see. I didn't understand why you weren't reusing `err`

jonboulle: Stop looking over my shoulder

dgonyeo: Sorry!

dgonyeo: And the type returned by runACBuild used to be a struct that implemented the error interface, and I was getting some weirdness at some point with that. That's not the case anymore, so this should go back to using `err`. My bad.

lib/copy.go
dgonyeo: The source in this situation could still be a directory.

jonboulle: ok, the option help text suggests otherwise

dgonyeo:
Did the things you wanted, also modified copy's new documentation to reflect these changes.

lib/copy.go
```diff
@@ -23,9 +24,15 @@ import (
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/coreos/rkt/pkg/uid"
 )
 
-// Copy will copy the directory/file at from to the path to inside the untarred
-// ACI at a.CurrentACIPath.
-func (a *ACBuild) Copy(from, to string) (err error) {
+// Copy has two modes of operation, selected with the explicitTarget flag.
```
jonboulle: OK. Now I'm going to annoy you by suggesting something I should have earlier, but only just occurred to me.

This should be two different functions for the two different codepaths. The `explicitTarget` conditional should be handled in the CLI code instead.

dgonyeo: Done

tests/dependency_test.go
```diff
@@ -72,9 +72,9 @@ func TestAddDependencyWithImageID(t *testing.T) {
 		t.Fatalf("%v\n", err)
 	}
 
-	hash, err1 := types.NewHash(depImageID)
-	if err1 != nil {
-		panic(err1)
+	hash, err := types.NewHash(depImageID)
```
jonboulle: By the by, a `types.MustNewHash` patch would be welcome

jonboulle:
Nice. Looks great now.

PR #104, opened 2015-11-12
fsdiffer: imported as subtree

dgonyeo:
Imported fsdiffer from https://github.com/dgonyeo/fsdiffer as a subtree
into util/fsdiffer


jonboulle:
@sgotti shall we assign you copyright here? it'd be dual as we've now patched it:
```
// Copyright Simone Gotti 2015
// Copyright CoreOS, Inc 2015
```

util/fsdiffer/.gitignore
```diff
@@ -0,0 +1,6 @@
+[._]*.s[a-w][a-z]
```
jonboulle: I think you can omit this now

dgonyeo:
For what it's worth, as far as dual copyright goes I didn't actually modify any of his files. Just added a couple.

jonboulle:
Couple of comments. Also, can you add copyright headers?  LGTM after that.

PR #105, opened 2015-11-12
util: removed Exists function

dgonyeo:
Didn't save a significant amount of code or logic.

### DIFF
--- a/lib/begin.go
+++ b/lib/begin.go
@@ -40,11 +40,13 @@ var (
 // an empty ACI, otherwise the ACI stored at start will be used at the starting
 // point.
 func (a *ACBuild) Begin(start string, insecure bool) (err error) {
-	ex, err := util.Exists(a.ContextPath)
-	if err != nil {
+	_, err = os.Stat(a.ContextPath)
+	switch {
+	case os.IsNotExist(err):
+		break
+	case err != nil:
 		return err
-	}
-	if ex {
+	default:
 		return fmt.Errorf("build already in progress in this working dir")
 	}
 
--- a/lib/common.go
+++ b/lib/common.go
@@ -21,15 +21,17 @@ import (
 	"syscall"
 
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema/types"
-
-	"github.com/appc/acbuild/util"
 )
 
 const defaultWorkPath = ".acbuild"
 
-// ErrNotFound is returned when acbuild is asked to remove an element from a
-// list and the element is not present in the list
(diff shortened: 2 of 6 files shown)


lib/begin.go
```diff
@@ -40,11 +40,13 @@ var (
...
-	if err != nil {
+	_, err = os.Stat(a.ContextPath)
+	switch {
+	case os.IsNotExist(err):
+		break
+	case err != nil:
 		return err
-	}
-	if ex {
+	default:
```
yifan-gu: Here if  `os.IsNotExist(err) == true`, the error will be returned by `Begin()`

How about just
```go
if err == nil {
    return fmt.Errorf("build already in progress in this working dir")
}
if !os.IsNotExist(err) {
    return err
}
return nil
```

yifan-gu: nvm.. didn't expand the function ..

lib/common.go
```diff
@@ -73,12 +71,12 @@ func NewACBuild(cwd string, debug bool) *ACBuild {
...
 func (a *ACBuild) lock() error {
-	ex, err := util.Exists(a.ContextPath)
-	if err != nil {
-		return err
-	}
-	if !ex {
+	_, err := os.Stat(a.ContextPath)
+	switch {
+	case os.IsNotExist(err):
 		return fmt.Errorf("build not in progress in this working dir - try \"acbuild begin\"")
```
yifan-gu: `build not in progress` or `no build in progress`?

dgonyeo: I think I like the phrasing in `no build in progress` more.

lib/end.go
```diff
@@ -17,19 +17,17 @@ package lib
...
-	ok, err := util.Exists(a.ContextPath)
-	if err != nil {
-		return err
-	}
-	if !ok {
+	_, err := os.Stat(a.ContextPath)
+	switch {
+	case os.IsNotExist(err):
 		return fmt.Errorf("build not in progress")
+	case err != nil:
```
yifan-gu: can we keep the error message consistent? 
maybe use a variable?
```go
 errorNoBuildInProgress = fmt.Errorf(`"build not in progress in this working dir - try \"acbuild begin\""`)
```

dgonyeo: Done.

yifan-gu:
LGTM

lib/begin.go
```diff
@@ -40,11 +40,13 @@ var (
...
 		return fmt.Errorf("build already in progress in this working dir")
```
jonboulle: For symmetry, this could be errBuildInProgress

Issue #111, opened 2015-11-17
systemd-nspawn: prints timezone warning


dgonyeo:
Every time I've seen `acbuild run` used, the following warning is printed before the actual command is exec'd:

```
Timezone Europe/Berlin does not exist in container, not updating container timezone.
```

We should prevent the error from being printed.

jonboulle:
@dgonyeo https://github.com/coreos/rkt/pull/226

PR #112, opened 2015-11-17
run: mirror host's /etc/localtime symlink target in rootfs

dgonyeo:
To prevent systemd from printing:

/etc/localtime does not point into /usr/share/zoneinfo/, not updating
container timezone.

The file pointed to by the /etc/localtime is copied into the expanded
ACI before execing systemd-nspawn.

Fixes https://github.com/appc/acbuild/issues/111 on any non-nixos system.

### DIFF
--- a/lib/run.go
+++ b/lib/run.go
@@ -17,9 +17,11 @@ package lib
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


jonboulle:
LGTM

rkusa:
Maybe this is related, I was trying out the nginx example using Vagrant and got `run: readlink /etc/localtime: invalid argument`

Full output:

```
vagrant@vagrant-ubuntu-vivid-64:~$ sudo ./build-nginx.sh 
Beginning build with an empty ACI
Setting name of ACI to example.com/nginx
Adding dependency "quay.io/coreos/alpine-sh"
Running: [apk update]
Downloading quay.io/coreos/alpine-sh: [========================] 2.65 MB/2.65 MB
run: readlink /etc/localtime: invalid argument
Ending the build
```

PR #118, opened 2015-11-18
copy: made -T the default behavior, put old default behind --to-dir

dgonyeo:
To make the semantics a little less confusing, I restored the original
default behavior of copy to be the default, and put the new behavior
introduced recently behind the `--to-dir` flag.

```
acbuild copy nginx.conf /etc/nginx/nginx.conf
acbuild copy --to-dir foo bar/* foo/bar /tmp/baz
```

### DIFF
--- a/acbuild/copy.go
+++ b/acbuild/copy.go
@@ -19,18 +19,18 @@ import (
 )
 
 var (
-	explicitTarget bool
-	cmdCopy        = &cobra.Command{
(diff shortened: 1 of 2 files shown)


dgonyeo:
v0.2.1 will go out once this is merged

jonboulle:
ok, LGTM, we can roll forward if we come up with something better.

iaguis:
:+1:

PR #126, opened 2015-11-19
modify: don't require . or / for starting char, don't delete on failure

dgonyeo:
When the modify flag was used, it was requiring that the initial
character for the path provided start with a `.` or a `/`. Since the
modify flag can only be used on local images, this was unnecessary.

Also when using the modify flag and the write failed (e.g. due to
permissions issues), the original image was deleted. This commit changes
it so that the write happens to a different file from the original, and
on success replaces the original with the new image.

### DIFF
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -171,28 +171,43 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd 
 
 		a := newACBuild()
 
-		err = a.Begin(aciToModify, false)
+		err = a.Begin(absoluteAciToModify, false)
 		if err != nil {
 			stderr("%v", err)
 			cmdExitCode = getErrorCode(err)
 			return
 		}
+
+		defer func() {
+			err = a.End()
+			if err != nil {
+				stderr("%v", err)
+				if cmdExitCode == 0 {
+					cmdExitCode = getErrorCode(err)
+				}
+			}
+		}()
 
 		cmdExitCode = cf(cmd, args)
 
-		err = a.Write(aciToModify, true, false, nil)
+		dir, file := path.Split(aciToModify)
+		tmpACIFile := path.Join(dir, "."+file+".tmp")
+
+		err = a.Write(tmpACIFile, true, false, nil)
 		if err != nil {
 			stderr("%v", err)
 			cmdExitCode = getErrorCode(err)
 			return
 		}
 
-		err = a.End()
+		err = os.Rename(tmpACIFile, aciToModify)
 		if err != nil {
+			os.Remove(tmpACIFile)
 			stderr("%v", err)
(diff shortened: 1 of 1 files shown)


acbuild/acbuild.go
```diff
@@ -131,6 +131,16 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
 			return
 		}
 
+		if len(aciToModify) == 0 {
```
jonboulle: how could this ever happen? probably need to panic (see 118)

acbuild/acbuild.go
```diff
@@ -131,6 +131,16 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
...
+			stderr("Path to ACI to modify can't be empty.")
+			cmdExitCode = 1
+			return
+		}
+
+		if aciToModify[0] != '.' && aciToModify[0] != '/' {
+			aciToModify = "./" + aciToModify
```
jonboulle: wait why is this necessary?

jonboulle: oh for begin? hmm. it's a little awkward

jonboulle: How about passing absoluteAciToMoidyf to Begin() instead

acbuild/acbuild.go
```diff
@@ -178,21 +188,36 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
...
+					cmdExitCode = getErrorCode(err)
+				}
+			}
+		}()
+
 		cmdExitCode = cf(cmd, args)
 
-		err = a.Write(aciToModify, true, false, nil)
+		dir, file := path.Split(aciToModify)
+		tmpACIFile := path.Join(dir, "."+file+".tmp")
```
jonboulle: any reason to roll your own tempfile here rather than using ioutil.Tempfile() or similar?

jonboulle: I guess it's ok, nvm.

acbuild/acbuild.go
dgonyeo: Yeah, for begin:

```
derek@energia ~/aci> acbuild --modify alpine-latest-linux-amd64.aci set-name dgretg
unknown host when fetching image, check your connection and local file paths must start with '/' or '.'
```

Passing `absoluteAciToModify` works too, I'll do that.

acbuild/acbuild.go
dgonyeo: Turns out the first thing I do in this function is `if aciToModify == ""`, so this check is redundant. I'll remove it.

dgonyeo:
Anything else you'd like me to tweak before I merge this?

jonboulle:
lgtm

Issue #132, opened 2015-11-29
acbuild run: readlink /etc/localtime: invalid argument


asbachb:
Hi,

I try to build the nginx example but script fails. Unsure if it's really an issue of acbuild or related to the alpine base image...

```
vm@ubuntu:~/acbuild/examples/nginx$ acbuild version
acbuild version v0.2.1-6-gbb2640c
appc version 0.7.1+git

vm@ubuntu:~/acbuild/examples/nginx$ sudo sh -x build-nginx.sh 
+ set -e
+ [  -ne 0 ]
build-nginx.sh: 4: [: Illegal number: 
+ acbuild --debug begin
Beginning build with an empty ACI
+ trap acbuildEnd EXIT
+ acbuild --debug set-name example.com/nginx
Setting name of ACI to example.com/nginx
+ acbuild --debug dep add quay.io/coreos/alpine-sh
Adding dependency "quay.io/coreos/alpine-sh"
+ acbuild --debug run apk update
Running: [apk update]
Downloading quay.io/coreos/alpine-sh: [========================] 2.65 MB/2.65 MB
run: readlink /etc/localtime: invalid argument
+ acbuildEnd
+ export EXIT=1
+ acbuild --debug end
Ending the build
+ exit 1
```

namsral:
I can replicate this issue on Ubuntu but not on CoreOS; on CoreOS the nginx example builds fine.

asbachb:
I think the problem is that /etc/localtime is not a symlink on Ubuntu:

```
vm@ubuntu:/tmp$ ls -l /etc/localtime 
-rw-r--r-- 1 root root 2335 Oct 23 17:15 /etc/localtime
vm@ubuntu:/tmp$ readlink /etc/localtime
vm@ubuntu:/tmp$ 
```

It would be nice to now what's the result of these commands on your CoreOS installation.

namsral:
CoreOS stable (835.7.0)

    $ readlink /etc/localtime
    ../usr/share/zoneinfo/UTC

asbachb:
The feature seems to be new:

https://github.com/appc/acbuild/commit/615e5cae57e0b0bcebf4498a4c561770f89bc16f

jonboulle:
I'm perplexed as to why this has never bitten us in rkt, since it follows basically the same codepath..

iaguis:
> I'm perplexed as to why this has never bitten us in rkt, since it follows basically the same codepath..

[...]

PR #134, opened 2015-11-30
run: handle /etc/localtime not being a symlink

dgonyeo:
Should fix https://github.com/appc/acbuild/issues/132

### DIFF
--- a/lib/run.go
+++ b/lib/run.go
@@ -266,18 +266,29 @@ func genDeplist(acipath string, reg registry.Registry) ([]string, error) {
 }
 
 func (a *ACBuild) mirrorLocalZoneInfo() error {
-	zif, err := os.Readlink("/etc/localtime")
-	if err != nil {
+	var localtimefile = "/etc/localtime"
+
+	finfo, err := os.Lstat(localtimefile)
+	switch {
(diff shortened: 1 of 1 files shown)


jonboulle:
I'm second-guessing this now: https://github.com/coreos/rkt/pull/226#discussion_r21409759
Perhaps we should just fix the readlink bug but not copy.

lib/run.go
```diff
@@ -266,18 +266,29 @@ func genDeplist(acipath string, reg registry.Registry) ([]string, error) {
...
-	if err != nil {
+	var localtimefile = "/etc/localtime"
+
+	finfo, err := os.Lstat(localtimefile)
+	switch {
+	case os.IsNotExist(err):
+		return nil
+	case err != nil:
 		return err
+	case finfo.Mode()|os.ModeSymlink > 0:
```
asbachb: I'll suggest that you do an & operation. I think this condition would also match when finfo.Mode() is something else than ModeSymlink.

```
case finfo.Mode() & os.ModeSymlink == os.ModeSymlink:
```

dgonyeo: Whoops, yeah I do want an `&`. Fixed.

Issue #131, opened 2015-11-27
systemd-nspawn: Failed to register machine: Machine 'target' already exists


dgonyeo:
systemd-nspawn determines the machine's name to enter in to machinectl via the target directory to chroot into. This directory is always named `target` by acbuild, thus every container acbuild starts up is called `target`. This means that multiple `acbuild run` invocations cannot be running at the same time.

`--machine acbuild-<hash-of-path>` flag should be passed to systemd-nspawn to prevent this.

jonboulle:
Why are these getting registered as machines? That doesn't seem right.

Derek Gonyeo <notifications@github.com> schrieb am Fr., 27. Nov. 2015 13:43:

dgonyeo:
That's the default.

```
--register=
           Controls whether the container is registered with systemd-machined(8). Takes a boolean argument, defaults to "yes". ...
```

It might make more sense to just disable registration though.

jonboulle:
Yep

Derek Gonyeo <notifications@github.com> schrieb am Fr., 27. Nov. 2015 15:14:

PR #135, opened 2015-11-30
run: don't register run commands with machined

dgonyeo:
This is necessary to allow for multiple run commands (for separate
builds) to be executing at the same time.

Fixes https://github.com/appc/acbuild/issues/131.


jonboulle:
to be extremely pedantic: the motivation for this change isn't just to facilitate concurrent invocations but also because it's not recommended practice for our use case - from `man 1 systemd-nspawn`:
> If the container does not run an init system, it is recommended to set this option to "no".

jonboulle:
LGTM

Issue #130, opened 2015-11-27
copy: copying to / produces error


ericchiang:
I have a tar of a root file system I'd like to use as my base when constructing an ACI. However the command `acbuild copy <src> /` produces an error.

```
$ mkdir rootfs
$ sudo tar -xv rootfs.tar -C rootfs
$ sudo acbuild begin
$ sudo acbuild copy rootfs /
copy: mkdir .acbuild/currentaci/rootfs: file exists
```

Current workaround is to remove `.acbuild/currentaci/rootfs`

```
$ sudo acbuild begin
$ sudo rm -rf .acbuild/currentaci/rootfs
$ sudo acbuild copy rootfs /
...
```

dgonyeo:
Yeah that's an issue. Shouldn't be hard to fix in acbuild, thanks for reporting this.

PR #136, opened 2015-11-30
begin: when given a directory, use as initial rootfs

dgonyeo:
To allow for users to specify an entire rootfs for their ACI, support
has been added to the begin command to allow for starting from a given
directory. This results in a non-empty rootfs, and an empty manifest.

Fixes https://github.com/appc/acbuild/issues/130, but with a different workflow than @ericchiang was expecting.

### DIFF
--- a/Documentation/subcommands/begin.md
+++ b/Documentation/subcommands/begin.md
@@ -11,7 +11,7 @@ the current directory is changed back to the location where `acbuild begin` was
 run. If this is undesirable, the `--work-path` flag can be provided to specify
 the location to store and access the build context.
 
-## Starting state
+## Starting with an empty ACI
 
 The build will default to starting with an empty ACI. The rootfs will be empty,
 and the manifest will look something like the following:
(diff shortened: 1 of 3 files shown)


Documentation/subcommands/begin.md
```diff
@@ -38,17 +38,33 @@ The `arch` and `os` labels are filled in with the architecture and operating
...
+begin. The contents of the directory will be copied into the ACI, and the ACI
+will have a manifest identical to when the begin command is not passed a
+directory.
+
+When specifying something on the local filesystem to the begin command, the
+path to it _must_ start with `.`, `~`, or `/`. As an example, if the directory
+is in the current directory, then instead of passing in `buildroot/output`,
+what would be passed in is `./buildroot/output`.
+
+## Starting state
```
jonboulle: Perhaps "Starting with a pre-existing ACI", and promote this above the rootfs one?

lib/begin.go
```diff
@@ -80,17 +82,59 @@ func (a *ACBuild) Begin(start string, insecure bool) (err error) {
...
 		} else {
 			return a.beginFromRemoteImage(start, insecure)
 		}
 	}
+	return a.beginWithEmptyACI()
+}
+
+func (a *ACBuild) beginFromLocalImage(start string) error {
+	finfo, err := os.Stat(start)
+	if err == nil {
```
jonboulle: this branching is slightly awkward, `if err != nil { return err }` seems more natural

jonboulle:
Couple nits but LGTM. Would be good to get an ack from @ericchiang

ericchiang:
LGTM

Issue #128, opened 2015-11-23
Support `aci-name:tag` syntax for both dep add and set-name


dgonyeo:
Such that these would be identical:

```
acbuild dep add example.com/nginx:5
acbuild dep add example.com/nginx --label version=5
```

```
acbuild set-name example.com/nginx:5
acbuild set-name example.com/nginx && acbuild label add version 5
```

PR #137, opened 2015-11-30
dependency: added shorthand for specifying version tag

dgonyeo:
Adds : syntax for specifying dependency versions. The two lines now do
the same thing:

```
acbuild dep add example.com/debian:jessie
acbuild dep add example.com/debian --label version=jessie
```

Partially fixes https://github.com/appc/acbuild/issues/128.

### DIFF
--- a/Documentation/subcommands/dependency.md
+++ b/Documentation/subcommands/dependency.md
@@ -36,6 +36,30 @@ The `add` command also has the following optional flags:
 - `--size`: the size of the dependency being added, in bytes. When this ACI is
   run, the retrieved dependency must have this size.
 
+The dependency being added also has a shorthand for specifying a version label
+with a `:` on the dependency's name. For example, the following two lines will
+behave identically:
+
+```bash
+acbuild dependency add example.com/debian --label version=jessie
+acbuild dependency add example.com/debian:jessie
+```
+
+## Image Name
+
+The `IMAGE_NAME` argument is a string that can define both the name used in
+meta discovery and a handful of labels. The syntax looks like this:
+`name[:version][,label1=value][,label2=value]`
+
+Everything except the name is optional. Some examples of this:
+
+```
+example.com/reduce-worker
+example.com/reduce-worker:1.0.0
+example.com/reduce-worker:1.0.0,channel=alpha
+example.com/reduce-worker:1.0.0,channel=alpha,env=staging
+```
+
 ## Examples
 
 ```bash
--- a/acbuild/dependency.go
+++ b/acbuild/dependency.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/discovery"
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/appc/spec/schema/types"
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
 )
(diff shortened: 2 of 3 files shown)


jonboulle:
can we use this? https://github.com/appc/spec/blob/master/discovery/parse.go#L45

dgonyeo:
sure

dgonyeo:
Uses `appc/spec/discovery.NewAppFromString` now. How's that look?

lib/dependencies.go
```diff
@@ -52,11 +55,23 @@ func (a *ACBuild) AddDependency(imageName, imageId string, labels types.Labels,
...
-	acid, err := types.NewACIdentifier(imageName)
+	app, err := discovery.NewAppFromString(imageName)
 	if err != nil {
 		return err
 	}
 
+	for name, value := range app.Labels {
+		for _, label := range labels {
+			if label.Name == name {
+				return fmt.Errorf("multiple %s tags specified", string(name))
```
jonboulle: is %s string(name) superfluous or is there a sublety I'm missing?

lib/dependencies.go
```diff
@@ -52,11 +55,23 @@ func (a *ACBuild) AddDependency(imageName, imageId string, labels types.Labels,
...
 	}()
 
-	acid, err := types.NewACIdentifier(imageName)
+	app, err := discovery.NewAppFromString(imageName)
 	if err != nil {
 		return err
 	}
 
+	for name, value := range app.Labels {
+		for _, label := range labels {
```
jonboulle: if _, ok := labels.Get(name); ok {

jonboulle:
a nit but otherwise lgtm

dgonyeo:
nits solved. Last call for more nits.

lib/dependencies.go
```diff
@@ -52,11 +55,21 @@ func (a *ACBuild) AddDependency(imageName, imageId string, labels types.Labels,
 		}
 	}()
 
-	acid, err := types.NewACIdentifier(imageName)
+	app, err := discovery.NewAppFromString(imageName)
```
jonboulle: Sorry for not catching this earlier, but I guess this is actually the wrong approach to use this here - it should be in acbuild/dependency.go instead. That way we can retain the actual signature of this function where imageName is really an imageName, and labels is a definitive set of labels. Make sense?

dgonyeo:
Yup makes sense. I just moved all of the string parsing and related sanity checking to `acbuild/dependency.go`. How's that look?

acbuild/dependency.go
```diff
@@ -71,8 +72,36 @@ func runAddDep(cmd *cobra.Command, args []string) (exit int) {
...
+	if err != nil {
+		stderr("dependency add: couldn't parse dependency name: %v", err)
+		return 1
+	}
+
+	appcLabels := types.Labels(labels)
+
+	for name, value := range app.Labels {
+		if _, ok := appcLabels.Get(string(name)); ok {
+			stderr("multiple %s tags specified", name)
```
jonboulle: s/tag/label/

jonboulle:
I guess the help for the add subcommand should be tweaked a bit too to reflect it's not strictly an image name..

dgonyeo:
Double last call for more nits.

jonboulle:
el gee tee emm

Documentation/subcommands/dependency.md
```diff
@@ -36,6 +36,30 @@ The `add` command also has the following optional flags:
 - `--size`: the size of the dependency being added, in bytes. When this ACI is
   run, the retrieved dependency must have this size.
 
+The dependency being added also has a shorthand for specifying a version label
+with a `:` on the dependency's name. For example, the following two lines will
```
blalor: The docs are out of sync with the implementation.

Issue #140, opened 2015-12-04
acbuild run with systemd-208 fails


blalor:
systemd v208 on CentOS 7.  `acbuild run` fails because this version of `systemd-nspawn` doesn't support the `-q/--quiet` option.

dgonyeo:
That's definitely an issue. I just made a PR that should fix this (https://github.com/appc/acbuild/pull/141), would you be willing to take it for a spin and make sure that this is fixed on your end?

blalor:
@dgonyeo different error, now. :-) 

    systemd-nspawn: unrecognized option '--register=no'

It appears `--register` was also [added in v209](https://github.com/systemd/systemd/commit/eb91eb187b7491e05fb95215b77cb62061f41d08).

blalor:
So, removing `--register` does allow `acbuild run`.  Assuming the behavior in v208 *is* to register the container with machined, I don't know what the impact of that is.

ecnahc515:
@blalor registration isn't really anything important, https://github.com/appc/acbuild/pull/135 removed registration, but it needs to check for the systemd version to ensure that flag is valid to fix this issue. 

I wonder if using nspawn is more effort to maintain than it's worth. Perhaps a pure go implementation for execution would make more sense (something like libcontainer with minimal options perhaps?). Or even running rkt (which I believe bundles systemd).

dgonyeo:
I'd prefer to just get around to https://github.com/appc/acbuild/issues/13, which pretty much addresses this (just with more work than dropping nspawn completely).

dgonyeo:
Just updated the PR for this issue to also not pass `--register` for when systemd < 209.

PR #141, opened 2015-12-04
run: don't pass -q flag to systemd-nspawn for systemd < 209

dgonyeo:
The -q/--quiet flag was added to systemd-nspawn in version 209, and some
distros (like CentOS 7) ship with an older version than that. This
commit runs and parses `systemctl --version` to determine the version
number of systemd, and will only pass the --quiet flag if it's high
enough.

Fixes https://github.com/appc/acbuild/issues/140

### DIFF
--- a/lib/run.go
+++ b/lib/run.go
@@ -120,7 +120,15 @@ func (a *ACBuild) Run(cmd []string, insecure bool) (err error) {
 
 		nspawnpath = a.OverlayTargetPath
 	}
-	nspawncmd := []string{"systemd-nspawn", "-q", "--register=no", "-D", nspawnpath}
+	nspawncmd := []string{"systemd-nspawn", "-D", nspawnpath}
+
+	version, err := getSystemdVersion()
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


lib/run.go
```diff
@@ -296,3 +305,25 @@ func (a *ACBuild) mirrorLocalZoneInfo() error {
...
+	if err == exec.ErrNotFound {
+		return 0, fmt.Errorf("system does not have systemd")
+	}
+
+	blob, err := exec.Command("systemctl", "--version").Output()
+	if err != nil {
+		return 0, err
+	}
+	for _, line := range strings.Split(string(blob), "\n") {
+		if strings.HasPrefix(line, "systemd ") {
```
jonboulle: ick, scanf seems saner

dgonyeo:
Uses scanf now, and also refrains from passing the `--register` flag for versions of systemd below 209.

jonboulle:
sure, lgtm if it works

blalor:
I tested this and it works for me.

PR #143, opened 2015-12-07
port: fix swapped protocol and port in add usage example

vreon:
The arguments for `acbuild port add` are `NAME`, `PROTOCOL`, and `PORT`,
but the example provided the latter two in the wrong order.


jonboulle:
LGTM, thanks!

Issue #120, opened 2015-11-18
*: generate annotations listing acbuild steps


dgonyeo:
It would be neat if acbuild kept track of the acbuild commands that get run on an ACI in the image's annotations.

It would provide inside into how a given image gets built.

PR #146, opened 2015-12-07
*: add annotations to track acbuild commands

dgonyeo:
This commit has acbuild maintain annotations following
"acbuild-command-%d". Each new acbuild command that isn't something like
cat-manifest or write results in a new annotation being added to the
image, containing the command that was run.

Fixes https://github.com/appc/acbuild/issues/120.

### DIFF
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -28,6 +28,7 @@ import (
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
 
 	"github.com/appc/acbuild/lib"
+	"github.com/appc/acbuild/util"
 )
 
 const (
@@ -63,9 +64,10 @@ OPTIONS:
 )
 
 var (
-	debug       bool
-	contextpath string
-	aciToModify string
+	debug          bool
+	contextpath    string
+	aciToModify    string
+	disableHistory bool
 
 	cmdExitCode int
 
@@ -91,6 +93,7 @@ func init() {
 	cmdAcbuild.PersistentFlags().BoolVar(&debug, "debug", false, "Print out debug information to stderr")
 	cmdAcbuild.PersistentFlags().StringVar(&contextpath, "work-path", ".", "Path to place working files in")
 	cmdAcbuild.PersistentFlags().StringVar(&aciToModify, "modify", "", "Path to an ACI to modify (ignores build context)")
+	cmdAcbuild.PersistentFlags().BoolVar(&disableHistory, "no-history", false, "Don't add annotations with the command that was run")
 
 	cobra.EnablePrefixMatching = true
 }
@@ -117,16 +120,27 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd 
 	return func(cmd *cobra.Command, args []string) {
 		if aciToModify == "" {
 			cmdExitCode = cf(cmd, args)
+			switch cmd.Name() {
+			case "cat-manifest", "begin", "write", "end", "version":
+				return
+			}
+			if cmdExitCode == 0 && !disableHistory {
+				err := addACBuildAnnotation(cmd, args)
+				if err != nil {
+					stderr("%v", err)
+					cmdExitCode = 1
+					return
+				}
+			}
 			return
 		}
 
-		command := strings.Split(cmd.Use, " ")[0]
-		switch command {
+		switch cmd.Name() {
 		case "cat-manifest":
 			cmdExitCode = runCatOnACI(aciToModify)
 			return
 		case "begin", "write", "end", "version":
-			stderr("Can't use the --modify flag with %s.", command)
+			stderr("Can't use the --modify flag with %s.", cmd.Name())
 			cmdExitCode = 1
 			return
 		}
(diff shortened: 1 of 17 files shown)


acbuild/acbuild.go
```diff
@@ -239,3 +262,30 @@ func stdout(format string, a ...interface{}) {
 	out := fmt.Sprintf(format, a...)
 	fmt.Fprintln(os.Stdout, strings.TrimSuffix(out, "\n"))
 }
+
+func addACBuildAnnotation(a *lib.ACBuild, cmd *cobra.Command, args []string) error {
+	if a == nil {
```
jonboulle: This isn't getting called with non-nil anywhere, what's that about?

acbuild/acbuild.go
```diff
@@ -239,3 +262,30 @@ func stdout(format string, a ...interface{}) {
...
+		var tmpCount int
+		n, _ := fmt.Sscanf(string(ann.Name), "acbuild-command-%d", &tmpCount)
+		if n == 1 && tmpCount > acbuildCount {
+			acbuildCount = tmpCount
+		}
+	}
+
+	command := cmd.Name()
+	for _, a := range args {
+		command += fmt.Sprintf(" %q", a)
```
jonboulle: oh the elegance of golang

acbuild/acbuild.go
```diff
@@ -239,3 +262,30 @@ func stdout(format string, a ...interface{}) {
...
+	}
+
+	return a.AddAnnotation(fmt.Sprintf("acbuild-command-%d", acbuildCount+1), command)
```
jonboulle: These should actually be namespaces more like... appc.io/acbuild/command or something

jonboulle: c.f. https://github.com/appc/docker2aci/blob/8c09a57b3976772258aaf5bbc1ee2dc6687e55c9/lib/common/common.go#L29

acbuild/acbuild.go
```diff
@@ -91,6 +93,7 @@ func init() {
 	cmdAcbuild.PersistentFlags().BoolVar(&debug, "debug", false, "Print out debug information to stderr")
 	cmdAcbuild.PersistentFlags().StringVar(&contextpath, "work-path", ".", "Path to place working files in")
 	cmdAcbuild.PersistentFlags().StringVar(&aciToModify, "modify", "", "Path to an ACI to modify (ignores build context)")
+	cmdAcbuild.PersistentFlags().BoolVar(&disableTracking, "no-tracking", false, "Don't add annotations with the command that was run")
```
jonboulle: Maybe "--no-history"? not sure. /cc @robszumski ?

acbuild/acbuild.go
dgonyeo: It gets called with nil here: https://github.com/appc/acbuild/pull/146/files#diff-003fdde72cd1c20a1cfa75450f56e198R128

acbuild/acbuild.go
dgonyeo: I like `--no-history` better than `--no-tracking` at the very least, so I switched to that.

acbuild/acbuild.go
jonboulle: **non**-nil

dgonyeo: Oh. I had intended to call it with non-nil when the `--modify` flag was used, but after reviewing a thing or two it's unnecessary. Removed the argument.

acbuild/acbuild.go
```diff
@@ -239,3 +262,28 @@ func stdout(format string, a ...interface{}) {
...
+
+	man, err := util.GetManifest(acb.CurrentACIPath)
+	if err != nil {
+		return err
+	}
+
+	var acbuildCount int
+	for _, ann := range man.Annotations {
+		var tmpCount int
+		n, _ := fmt.Sscanf(string(ann.Name), "appc.io/acbuild/command-%d", &tmpCount)
```
jonboulle: Hrm, split out the fmt string as a const please

tests/annotation_test.go
```diff
@@ -43,7 +43,7 @@ func TestAddAnnotation(t *testing.T) {
 	workingDir := setUpTest(t)
 	defer cleanUpTest(workingDir)
 
-	_, _, _, err := runACBuild(workingDir, "annotation", "add", annoName, annoValue)
+	_, _, _, err := runACBuild(workingDir, "--no-history", "annotation", "add", annoName, annoValue)
```
jonboulle: are you ever running this with history?!
since 99% of invocations seem to be of the pattern
`_, _, _, err := runACBuild(workingDIr, "--no-history", ...)`
might I suggest a helper func

dgonyeo:
nits should be fixed

acbuild/acbuild.go
dgonyeo: `let command = cmd.Name ++ (foldr (++) "" (map (fmt.Sprintf " %q") args))`

dgonyeo: Actually:
`let command = cmd.Name ++ (intercalate " " args)`

jonboulle:
Code LGTM. Please add a note to the README explaining the behaviour + that it can be disabled.

dgonyeo:
I added a file to the `Documentation` directory describing the new functionality. How's that look?

acbuild/acbuild.go
```diff
@@ -239,3 +262,39 @@ func stdout(format string, a ...interface{}) {
 	out := fmt.Sprintf(format, a...)
 	fmt.Fprintln(os.Stdout, strings.TrimSuffix(out, "\n"))
 }
+
+func addACBuildAnnotation(cmd *cobra.Command, args []string) error {
+	const annoNamePattern = "appc.io/acbuild/command-%d"
```
jonboulle: When I asked for this to be a const, my thinking was that it would be exported and essentially part of the acbuild API contract. But that might be premature, this is fine for now.

Issue #142, opened 2015-12-05
extra files in aci after run


blalor:
I'm attempting to use acbuild on CentOS 7 (systemd 208).  Running on a patched branch of #141 to address issues I reported in #140, I'm seeing extra files in the ACI aside from the mandatory `rootfs` and `manifest`.  

Steps to reproduce (the dependency should be publicly accessible for the time being):
```
[root@localhost ~]# /tmp/acbuild/bin/acbuild begin
[root@localhost ~]# /tmp/acbuild/bin/acbuild dependency add s3.amazonaws.com/bsd-aci/base/centos7 --label version=4b6264c-14
[root@localhost ~]# /tmp/acbuild/bin/acbuild mount add resolvconf /etc/resolv.conf --read-only
[root@localhost ~]# /tmp/acbuild/bin/acbuild set-name whatev
[root@localhost ~]# /tmp/acbuild/bin/acbuild set-exec /bin/bash
[root@localhost ~]# /tmp/acbuild/bin/acbuild run /bin/true
Downloading s3.amazonaws.com/bsd-aci/base/centos7: [===========] 47 MB/47 MB  
[root@localhost ~]# /tmp/acbuild/bin/acbuild write foo.aci
[root@localhost ~]# tar -tzvf foo.aci 
drwxr-xr-x 0/0               0 2015-12-05 04:03 rootfs
drwxr-xr-x 0/0               0 2015-12-05 04:04 rootfs/etc
lrwxrwxrwx 0/0               0 2015-12-05 04:04 rootfs/etc/localtime -> ../usr/share/zoneinfo/Etc/UTC
-rw-r--r-- 0/0               0 2015-12-05 04:04 rootfs/etc/resolv.conf
drwxr-xr-x 0/0               0 2015-12-05 04:04 usr
drwxr-xr-x 0/0               0 2015-12-05 04:04 usr/share
drwxr-xr-x 0/0               0 2015-12-05 04:04 usr/share/zoneinfo
drwxr-xr-x 0/0               0 2015-12-05 04:04 usr/share/zoneinfo/Etc
-rw-r--r-- 0/0             118 2015-12-05 04:04 usr/share/zoneinfo/Etc/UTC
-rw-r--r-- root/root       393 2015-12-05 04:04 manifest
```

[...]

PR #147, opened 2015-12-07
Fully resolves the path to /etc/localtime on the host.

### LINKED ISSUE #145: relative symlink for /etc/localtime causes run to fail
host machine: `readlink /etc/localtime -> ../usr/share/zoneinfo/Etc/UTC`

setup:

```
acbuild begin
acbuild dependency add s3.amazonaws.com/bsd-aci/base/centos7 --label version=4b6264c-14
acbuild mount add resolvconf /etc/resolv.conf --read-only
acbuild set-name whatev
acbuild set-exec /bin/bash
acbuild run /bin/true
```

If I run this in `/tmp` it works.  If I run in `/tmp/foo` it fails because `acbuild` resolves the symlink relative to the current directory, rather than to `/etc`:
```
run: open ../usr/share/zoneinfo/Etc/UTC: no such file or directory
```

blalor:
Fixes #145.  Also affects #142: `usr` is no longer a child of `rootfs` and `manifest` in `.acbuild/currentaci`.

### DIFF
--- a/lib/run.go
+++ b/lib/run.go
@@ -266,7 +266,7 @@ func genDeplist(acipath string, reg registry.Registry) ([]string, error) {
 }
 
 func (a *ACBuild) mirrorLocalZoneInfo() error {
-	zif, err := os.Readlink("/etc/localtime")
+	zif, err := filepath.EvalSymlinks("/etc/localtime")
 	if err != nil {
 		return err
 	}


dgonyeo:
Related: https://github.com/appc/acbuild/pull/134. I'd be fine with rebasing that on top of this.

dgonyeo:
LGTM, cc @jonboulle

blalor:
#134 may actually be obsoleted by this. [`EvalSymlinks`](https://golang.org/pkg/path/filepath/#EvalSymlinks):

> EvalSymlinks returns the path name after the evaluation of any symbolic links. If path is relative the result will be relative to the current directory, unless one of the components is an absolute symbolic link.

lib/run.go
```diff
@@ -266,7 +266,7 @@ func genDeplist(acipath string, reg registry.Registry) ([]string, error) {
 }
 
 func (a *ACBuild) mirrorLocalZoneInfo() error {
-	zif, err := os.Readlink("/etc/localtime")
+	zif, err := filepath.EvalSymlinks("/etc/localtime")
```
jonboulle: Is this preferable to the rkt approach for some reason? https://github.com/coreos/rkt/blob/master/stage1/init/init.go#L106-L110
if so we should probably switch there too

jonboulle:
yeah per https://github.com/appc/acbuild/pull/134#issuecomment-160739791 I think I'd rather axe that and just go with this fix.

lib/run.go
jonboulle: @dgonyeo ?

Issue #139, opened 2015-12-01
Error extracting tarball when using debian docker base


ecnahc515:
```
vagrant@vagrant-ubuntu-vivid-64:/tmp$ docker2aci docker://debian:sid
Downloading 1a8ce871807c: [====================================] 51.6 MB/51.6 MB
Downloading 00399b985516: [====================================] 32 B/32 B

Generated ACI(s):
debian-sid.aci
vagrant@vagrant-ubuntu-vivid-64:/tmp$ acbuild begin ./debian-sid.aci
begin: error extracting tarball: link /rootfs/bin/gunzip /tmp/.acbuild/currentaci/rootfs/bin/uncompress: no such file or directory
```

iaguis:
+1

dgonyeo:
Pretty sure I know what's causing this. As of https://github.com/appc/acbuild/commit/244d7bb6dd0ef914a37eb8b79c3c642a768a7b3d `acbuild begin` no longer requires root. When running as root it'll chroot into the target and extract things from there, and when running as an unprivileged user it'll just extract things into the target without the chroot. Turns out `/rootfs/bin/gunzip` isn't a valid file to make a link to when it's not in the chroot. I'll make a PR for rkt's tar library and then bump the version acbuild pulls in.

ecnahc515:
@dgonyeo Not closed, fix needs to be applied to acbuild.

PR #148, opened 2015-12-08
rkt: bumped vendored version to latest, pulls in fix for tar package

dgonyeo:
The tar package didn't correctly handle hard links when the
ExtractTarInsecure package was called directly. This was recently fixed
in rkt, so this commit it to bump the version of rkt that acbuild uses.

Fixes https://github.com/appc/acbuild/issues/139 (for real this time).

cc @ecnahc515


jonboulle:
LGTM

PR #152, opened 2015-12-18
run: set the PATH variable if not set by manifest

dgonyeo:
Depending on the image, the PATH variable inside of a acbuild run
command could be propagated from something on the host. On normal
distros, this was mostly a non-issue. On NixOS, it would get set to
"/no-such-path:/sbin:/bin", which is definitely an issue.

This commit checks if the PATH environment variable has been set in the
ACI's manifest (which gets added to the systemd-nspawn command), and if
it has not it sets it to
"/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin".

Note that it only sets it on the systemd-nspawn command, it doesn't
add or modify the PATH variable in the manifest.

### DIFF
--- a/lib/run.go
+++ b/lib/run.go
@@ -135,6 +135,7 @@ func (a *ACBuild) Run(cmd []string, insecure bool) (err error) {
 			nspawncmd = append(nspawncmd, "--setenv", evar.Name+"="+evar.Value)
 		}
 	}
+	nspawncmd = append(nspawncmd, "--setenv", "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
 
 	err = a.mirrorLocalZoneInfo()
 	if err != nil {


dgonyeo:
Oh, and I got the value for the variable from:
https://github.com/appc/spec/blob/master/spec/ace.md#execution-environment

lib/run.go
```diff
@@ -131,9 +131,16 @@ func (a *ACBuild) Run(cmd []string, insecure bool) (err error) {
 	}
 
 	if man.App != nil {
+		setPath := false
 		for _, evar := range man.App.Environment {
+			if evar.Name == "PATH" {
```
jonboulle: Strictly speaking you don't even need to bother with this right now, the spec mandates it's always this value: https://github.com/appc/spec/blame/master/spec/ace.md#L92

dgonyeo: How does rkt handle when a manifest sets a path variable then? Does it just ignore it?

jonboulle: Yes. Maybe room for improvement there. Feel free to file a bug/fix :-)

dgonyeo: Setting the `PATH` for acbuild will probably be more useful than setting the `PATH` for rkt, since the AppC spec forces you to provide an absolute path to the binary you're exec'ing, and `acbuild run` doesn't. I still think it'd be useful to override the default in rkt though, since I could probably come up with some edge case related to `nix2aci` where having a weird `PATH` would be useful.

I think rkt should do what I'm doing here. I'll file an issue against rkt to see what people think.

dgonyeo:
Regardless of the rkt issue I just made, thoughts on me merging this? I could set it to what the AppC Spec dictates until the issue I made gets resolved, if that'd be better.

This is annoying:
```
derek@rokot ~/go/src/github.com/appc/acbuild> sudo acbuild begin                         
derek@rokot ~/go/src/github.com/appc/acbuild> sudo acbuild dep add aci.gonyeo.com/debian 
derek@rokot ~/go/src/github.com/appc/acbuild> sudo acbuild run -- /bin/sh -c 'echo $PATH'
Downloading aci.gonyeo.com/debian: [===========================] 37.2 MB/37.2 MB
/no-such-path:/sbin:/bin
```

lib/run.go
dgonyeo: Oh, apparently the requirement for an absolute path for the exec field changed recently.

jonboulle: Can you remove this check for now?

jonboulle:
Yes, please just do per the comments/appc spec for now.

dgonyeo:
Updated to always set the `PATH` variable as per the spec. I'll look into making a PR to change the spec, as I think this should be an overridable thing.

jonboulle:
lgtm

PR #158, opened 2015-12-22
examples: added java7, which uses the host's apt-get - WIP

dgonyeo:
Putting this up for review. Rewrote the java7 section of https://github.com/appc/build-repository to use acbuild.

Addresses some (maybe all?) of https://github.com/appc/build-repository/issues/10.

cc @jonboulle


examples/java7/build-java7.sh
```diff
@@ -0,0 +1,82 @@
...
+PACKAGESDIR=$(mktemp -d)
+
+rmDirectories() {
+    rm -rf ${TMPDIR}
+    rm -rf ${PACKAGESDIR}
+}
+trap rmDirectories EXIT
+
+pushd ${PACKAGESDIR}
+apt-get download \
```
jonboulle: holy giant unordered list of packages batman.

examples/java7/build-java7.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
+
+set -e
+
+which acbuild || (echo "Make sure acbuild is on your path"; exit 1)
```
jonboulle: which apt-get maybe?

jonboulle:
Sure, if it works lgtm

examples/java7/build-java7.sh
dgonyeo: Well it came from the giant unordered list of packages that was a separate file before. Think if I kept this in a separate file still it would be a little saner?

jonboulle: no strong opinion, do what you think is best

jonboulle:
ping

dgonyeo:
Needs testing. Will test next week.

examples/java7/build-java7.sh
jellonek: IMO this should go into separate .txt file which then could be used like `$(cat deb-requirements.txt)` - any further modification of this list would not touch this file.

examples/java7/build-java7.sh
```diff
@@ -0,0 +1,82 @@
...
+    zlib1g \
+    bash \
+    libncurses5 \
+    libtinfo5
+popd
+
+shopt -s nullglob
+for d in ${PACKAGESDIR}/*.deb; do
+  echo "Extracting $d"
+  dpkg-deb -x ${d} ${TMPDIR}
```
jellonek: Doing that - You are skipping postinst process, with also skip of post install triggers (ldconfig for example).
Maybe using debootstrap with `--include` parameter would be more secure option for doing above work?

dgonyeo: I tried using debootstrap to make a base, and without installing anything on top of that the ACI is at 134 MB.

I tried using the instructions in [this SO post](https://stackoverflow.com/questions/3954584/running-apt-get-for-another-partition-directory) to just `apt-get install` nginx into a different root dir. It pulled in so many packages (packages like apt, adduser, mount, e2fslibs, ...) that it would use up 163 MB of disk space, and ended up erroring on the install process when `/usr/bin/apt-listchanges` wasn't found.

Just extracting deb files into a directory skips potentially important install steps, and doing it correctly signs you up for containers greater in size than 134 MB. I really think the only decent solution to integrating a package manager with a container build system is nix2aci, where you have a package manager that doesn't need to include everything and the kitchen sink in case there are undeclared dependencies (nix is guaranteed to actually know the dependencies of its packages, instead of "oh the guy had this binary available when he built the package so you need to too").

So @jonboulle, want me to leave this in its current state, make the containers massive, or drop this and just point to nix2aci?

jonboulle: Can you replace this with a nix java7 example?

dgonyeo: Certainly

dgonyeo:
Closing in favor of https://github.com/appc/acbuild/pull/158#discussion_r53969283.

PR #166, opened 2016-02-10
Fix set-name example text.

rillian:
The example gives `acbuild name quay.io/foo/bar` which doesn't work.


dgonyeo:
Oh my, I completely missed. Thanks for the PR!

Issue #168, opened 2016-02-10
Recovering from pre-terminated build: device or resource busy


zopyx:
I tried to remove .acbuild because of a pre-terminated build due to a broken SSH connection. This fails:

[root@vmd9666 plone-aci2]#  rm -fr .acbuild/
rm: cannot remove '.acbuild/target': Device or resource busy

Checking the open file handles using lsof did not reveal any process holding a lock on the .acbuild directory.

alban:
Is there any mount points in this directory? Or is `systemd-nspawn` still running in there? Its current working directory might be in there.

zopyx:
What do you mean with mount-points?

```
 [root@vmd9666 ~]# ps auxw|grep syste
root         1  0.0  0.0  43920  5260 ?        Ss   Jan23   1:31
/usr/lib/systemd/systemd --system --deserialize 18
root       439  0.0  1.6 201820 98800 ?        Ss   Jan23   4:38
/usr/lib/systemd/systemd-journald
dbus       560  0.0  0.0  54964  3408 ?        Ss   Jan23   0:43
/usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile
--systemd-activation
root       611  0.0  0.0  24844  3124 ?        Ss   Jan23   0:17
/usr/lib/systemd/systemd-logind
plone      633  0.0  0.0  45324  3764 ?        Ss   Jan23   0:03
/usr/lib/systemd/systemd --user
root      2401  0.0  0.0  43208  3872 ?        Ss   Jan26   0:02
/usr/lib/systemd/systemd --user
root     11199  0.0  0.0   9120   952 pts/0    S+   13:10   0:00 grep
--color=auto syste
root     13443  0.0  0.0  17908  2516 ?        Ss   Feb08   0:01
/usr/lib/systemd/systemd-machined
root     32295  0.0  0.0  43328  2876 ?        Ss   Feb08   0:00
/usr/lib/systemd/systemd-udevd
```

dgonyeo:
During a run command acbuild will mount an overlayfs to `.acbuild/target`. It would probably be reasonable for acbuild to see if this is already mounted and try to umount it if it is.

For specifically your case, see if running `umount .acbuild/target` from the directory you ran acbuild in will allow you to delete the `.acbuild` directory.

PR #170, opened 2016-02-10
run/end: check if .acbuild/target is mounted, unmount if so

dgonyeo:
If acbuild was killed in the middle of an `acbuild run` command, the
overlayfs it mounts at .acbuild/target would remain mounted. This caused
issues whenever the user attempts to use `acbuild end` to clean up after
the failed build.

Fixes https://github.com/appc/acbuild/issues/168.

### DIFF
--- a/lib/end.go
+++ b/lib/end.go
@@ -16,6 +16,8 @@ package lib
 
 import (
 	"os"
+
+	"github.com/appc/acbuild/util"
 )
 
 // End will stop the current build. An error will be returned if no build is in
@@ -30,6 +32,11 @@ func (a *ACBuild) End() error {
 	}
 
 	if err = a.lock(); err != nil {
+		return err
+	}
+
+	err = util.MaybeUnmount(a.OverlayTargetPath)
+	if err != nil {
 		return err
 	}
 
--- a/lib/run.go
+++ b/lib/run.go
@@ -54,6 +54,11 @@ func (a *ACBuild) Run(cmd []string, insecure bool) (err error) {
 		return fmt.Errorf("the run subcommand must be run as root")
 	}
 
+	err = util.MaybeUnmount(a.OverlayTargetPath)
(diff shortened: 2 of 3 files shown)


lib/end.go
```diff
@@ -16,12 +16,34 @@ package lib
...
+	"syscall"
+
+	"github.com/appc/acbuild/util"
 )
 
 // End will stop the current build. An error will be returned if no build is in
 // progress.
 func (a *ACBuild) End() error {
-	_, err := os.Stat(a.ContextPath)
+	_, err := os.Stat(a.OverlayTargetPath)
```
jonboulle: How about moving this all into a util.MaybeUnmount() function

dgonyeo:

lib/run.go
```diff
@@ -54,6 +54,25 @@ func (a *ACBuild) Run(cmd []string, insecure bool) (err error) {
...
 	}
 
+	_, err = os.Stat(a.OverlayTargetPath)
+	switch {
+	case os.IsNotExist(err):
+		break
+	case err != nil:
+		return err
+	default:
+		alreadyMounted, err := util.CheckIfMounted(a.OverlayTargetPath)
```
jonboulle: alreadyMounted? how about just mounted?

lib/run.go
```diff
@@ -54,6 +54,25 @@ func (a *ACBuild) Run(cmd []string, insecure bool) (err error) {
 		return fmt.Errorf("the run subcommand must be run as root")
 	}
 
+	_, err = os.Stat(a.OverlayTargetPath)
```
jonboulle: wait, confused, why isn't' this calling maybeunmount?

dgonyeo: Because I forgot to call it here. Fixed.

jonboulle:
lgtm

Issue #164, opened 2016-02-05
exit status 1 all the time?


cdrage:
For some reason I'm getting this when building all the examples:
```bash
acbuild/examples/apache  master ✗                                                                                                                                                                                                                      30d ◒  
▶ sudo ./build-apache.sh
Beginning build with an empty ACI
Setting name of ACI to example.com/apache
Adding dependency "quay.io/coreos/alpine-sh"
Running: [apk update]
run: exit status 1
Ending the build
```

Any idea as to why? No debug info for this :(

cdrage:
Using:
```
acbuild/examples/apache  master ✗                                                                                                                                                                                                                     30d ◒  ⍉
▶ acbuild version
acbuild version v0.2.2-2-g46ad9c4
appc version 0.7.1+git
```
(master)

dgonyeo:
Related: https://github.com/appc/acbuild/issues/162

I'm aware of this as a problem, but have no clue what's causing it. I'll look into it soon.

cdrage:
@dgonyeo #162 
Okay, so I think I **may** have found the problem.

If you build using:
`acbuild --debug begin quay.io/listhub/alpine`

it works fine (without adding dep)

but if you use it depending on the dep, it doesn't
`acbuild --debug dep add quay.io/coreos/alpine-sh` 

any ideas?

dgonyeo:
That may mean that there's some issue with overlayfs, I'll see if I can investigate further.

cdrage:
FYI, I'm building on Debian 8 with `3.16.0-4-amd64`.

dgonyeo:
I think I may have had a friend who saw a similar, if not identical issue also in Debian once. Guess it's time to spin up a Debian vm.

dgonyeo:
Figured it out: acbuild doesn't print a useful error message when the overlayfs kernel module isn't available.

At this line: https://github.com/appc/acbuild/blob/master/lib/run.go#L83 acbuild execs `modprobe overlay`, and if an error is encountered it's simply returned. A much better behavior would be to detect when the error is exit code 1, which implies that the module failed to load. I'll make a PR.

For specifically debian, I found [this](https://lists.debian.org/debian-live/2014/12/msg00041.html), which says that overlayfs didn't make it into Linux until 3.18. It also looks like you can get the newer kernel from experimental, so I'd suggest you give that a shot if you want to get acbuild working in debian.

[...]

PR #171, opened 2016-02-10
run: show a friendlier error message when unable to modprobe overlayfs

dgonyeo:
If overlayfs isn't available on a user's system, acbuild attempts to modprobe it. If this fails, acbuild would exit with `run: exit status 1`. This is a very confusing error message to receive, so this commit checks for this failure and instead prints `run: overlayfs is not supported on your system`.

Fixes https://github.com/appc/acbuild/issues/164

### DIFF
--- a/lib/run.go
+++ b/lib/run.go
@@ -82,6 +82,9 @@ func (a *ACBuild) Run(cmd []string, insecure bool) (err error) {
 		if !supportsOverlay() {
 			err := exec.Command("modprobe", "overlay").Run()
 			if err != nil {
+				if _, ok := err.(*exec.ExitError); ok {
+					return fmt.Errorf("overlayfs is not supported on your system")
+				}
 				return err
 			}
 			if !supportsOverlay() {


cdrage:
Thank you!

Confirm this errors out correctly. I knew it was something to do with overlay as I had been git loggin' back to previous builds and it actually errored out previously with overlay. 

```
dropbox/dev/acbuild  171-dgonyeo-better-overlay-error ✔                                                                                                                                                                                                  16m  
▶ ./build && sudo cp bin/acbuild /usr/bin/ 
Building acbuild...

dropbox/dev/acbuild  171-dgonyeo-better-overlay-error ✔                                                                                                                                                                                                  16m  
▶ ls
acbuild  bin  build  build-docker  CHANGELOG.md  Documentation  examples  Godeps  gopath  lib  LICENSE  README.md  registry  scripts  test  tests  util  Vagrantfile

dropbox/dev/acbuild  171-dgonyeo-better-overlay-error ✔                                                                                                                                                                                                  16m  
▶ bash examples/nginx/build-nginx.sh 
This script uses functionality which requires root privileges

dropbox/dev/acbuild  171-dgonyeo-better-overlay-error ✔                                                                                                                                                                                                 16m  ⍉
▶ sudo bash examples/nginx/build-nginx.sh
Beginning build with an empty ACI
Setting name of ACI to example.com/nginx
Adding dependency "quay.io/coreos/alpine-sh"
Running: [apk update]
run: overlayfs is not supported on your system
Ending the build
```

LGTM :+1:

jonboulle:
seems fine - do we want to do anything more nuanced? e.g. if modprobe isn't found?

cdrage:
@jonboulle 
tell the user to get their bum on kernel 3.18?

an alternative would be to build via `begin quay.io/listhub/alpine` instead of `dep add quay.io/coreos/alpine`? don't know if that'd be possible tho

dgonyeo:
I think a systemd without modprobe is a strong enough edge case where I don't care. Once I finish reverting the duplicate PR I accidentally merged, ok to merge this @jonboulle?

jonboulle:
OK

PR #174, opened 2016-02-16
Update README.md

n0rad:
update BlaBlaCar build's project to dgr


dgonyeo:
Just a heads up, the gif in dgr's readme still says "cnt".

Issue #163, opened 2016-02-01
can't copy to /


mischief:
copy seems to abort if rootfs already exists when the target is /.

```
mischief@delta /tmp $ mkdir a
mischief@delta /tmp $ acbuild begin
mischief@delta /tmp $ acbuild copy a /
copy: mkdir .acbuild/currentaci/rootfs: file exists
```

mischief:
seems --to-dir fixes this, but the UX here could be improved :-)

dgonyeo:
Yup, `--to-dir` is what you want.

The reasoning for the bizarre UX is that acbuild lazily fetches dependencies, and as such when you call `acbuild copy nginx.conf /etc/nginx`, acbuild doesn't know if the `/etc/nginx` directory exists. For all it knows you want the file `/etc/nginx` inside the container to be a copy of the file `nginx.conf` outside of the container. acbuild could fetch the dependencies for an `acbuild copy` command so it can know if directories exist, but downloading tens to hundreds of megabytes to copy a file into a tarball seems excessive. To make the problem worse, acbuild doesn't currently hold on to dependencies between runs.

One options is that once https://github.com/appc/acbuild/issues/125 is done acbuild could hold on to dependencies between runs, in which case preemptively fetching dependencies will be more reasonable. (I don't have too much faith in the current store logic).

Another option is to remove `acbuild copy` and replace it with `acbuild copy-to-dir` and `acbuild copy-file` or something similar, so the difference is a little more obvious than having it fail on you and needing to dig around for `acbuild copy`'s flags.

PR #177, opened 2016-02-17
copy: split --to-dir flag into separate subcommand

dgonyeo:
Copy's semantics are weird and I don't like them, but there are
technical reasons preventing `acbuild copy` from being reasonable. This
commit breaks the --to-dir flag on `acbuild copy` out into a separate
command, so that when users notice that `acbuild copy` doesn't do what
they expect it'll be a little more obvious that there are two modes of
operation here. Hiding this away in a flag, and therefore inside a help
page that wasn't at the top level, wasn't very user friendly.

Fixes https://github.com/appc/acbuild/issues/163.

### DIFF
--- a/Documentation/subcommands/copy-to-dir.md
+++ b/Documentation/subcommands/copy-to-dir.md
@@ -1,0 +1,22 @@
+# acbuild copy-to-dir
+
+`acbuild copy-to-dir` will copy any number of files and directories from the
+local filesystem into the ACI.
+
+It takes at least two arguments, where all but the final argument are paths on
+the host system. The final argument is a parent directory inside the ACI to
+place the files and directories in.
+
+If the target directory doesn't exist in the ACI, it will be implicitly created
+(along with any necessary parent directories).
+
+The following commands would do the same thing:
+
+```bash
+acbuild copy-to-dir apache.conf sites-available/00-default sites-available/myblog /etc/apache2
+```
+
+```bash
+cp apache.conf sites-available/00-default sites-available/myblog ./.acbuild/currentaci/rootfs/etc/apache2
+```
+
--- a/acbuild/copy-to-dir.go
+++ b/acbuild/copy-to-dir.go
@@ -15,40 +15,49 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
 )
 
 var (
-	cmdCopy = &cobra.Command{
-		Use:     "copy PATH_ON_HOST PATH_IN_ACI",
-		Short:   "Copy a file or directory into an ACI",
-		Example: "acbuild copy nginx.conf /etc/nginx/nginx.conf",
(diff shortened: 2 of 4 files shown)


Documentation/subcommands/copy-to-dir.md
```diff
@@ -0,0 +1,22 @@
+# acbuild copy-to-dir
+
+`acbuild copy` will copy any number of files and directories from the local
```
jonboulle: copy-to-dir

acbuild/copy-to-dir.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/appc/acbuild/Godeps/_workspace/src/github.com/spf13/cobra"
+)
+
+var (
+	toDir        bool
+	cmdCopyToDir = &cobra.Command{
+		Use:     "copy-to-dir PATH1_ON_HOST PATH2_ON_HOST ... PATH_IN_ACI",
+		Short:   "Copy a file or directory into a directory in an ACI",
+		Example: "acbuild copy build/bin/* /usr/bin",
```
jonboulle: update meeeee

acbuild/copy-to-dir.go
```diff
@@ -0,0 +1,64 @@
...
+		return 1
+	}
+	if len(args) < 2 {
+		stderr("copy-to-dir: incorrect number of arguments")
+		return 1
+	}
+
+	if debug {
+		logMsg := "Copying "
+		for i := 0; i < len(args)-1; i++ {
```
jonboulle: i := range args

acbuild/copy-to-dir.go
```diff
@@ -0,0 +1,64 @@
...
+			logMsg += fmt.Sprintf("host:%s ", args[i])
```
jonboulle: this format seems a little weird (or is there already precedent within acbuild?). not that I have a better suggestion. just peanut gallery.

jonboulle:
Seems OK to me. Would be great if we could get feedback somehow from those who struggled with the current UX.

dgonyeo:
@mischief: any thoughts on this?

acbuild/copy-to-dir.go
dgonyeo: Just tweaked it. Any better now?

dgonyeo:
Well I think this will help the UX, although I think what really needs to happen for copy is that acbuild preemptively fetches dependencies so it can know about the container's directory structure. In its current state, with `acbuild copy etcd /bin` acbuild can't know if `/bin` exists, since that might be in an unfetched dependency.

Just need to get a more permanent on disk store for that to happen, because always fetching dependencies no matter what becomes hell if acbuild doesn't hold on to them between builds.

Honestly this commit is just more of a short-term fix until I can actually fix copy.

dgonyeo:
Last call for nits before I merge this.

jonboulle:
Go for it.

Issue #160, opened 2016-01-15
acbuild exit code always 0


f0:
if a acbuild subcommand failed, the exit code is always 0, but should be > 0

e.g acbuild annotation a $b   <-- add is missed

comes back with $?=0,  so its not possible to check for problems in the build!

dgonyeo:
Yup this is definitely an issue. I've known about it for a little while, but I guess I never made an issue. 

That exit code is coming from cobra, so I'm not positive how to fix it. Worst case scenario we have to modify cobra, but maybe it exposes a way to set the exit code on a failed parse. 

Thanks for reporting this.

PR #178, opened 2016-02-18
*: exit with error code 3 when arguments are invalid

dgonyeo:
If the user calls acbuild with invalid subcommands or flags, and the
code for no subcommand is called, acbuild would up until now exit with
exit code 0. This can break scripts relying on the exit code to catch
errors.

This commit changes this such that acbuild will exit with code 3 when
called with invalid subcommands or flags.

Fixes https://github.com/appc/acbuild/issues/160.

### DIFF
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -71,6 +71,8 @@ var (
 
 	cmdExitCode int
 
+	errCobra = fmt.Errorf("cobra error")
+
 	templFuncs = template.FuncMap{
 		"subcmdList": func(cmds []*cobra.Command) string {
 			var subcmds []string
@@ -106,6 +108,8 @@ func getErrorCode(err error) int {
 	switch err {
 	case lib.ErrNotFound:
 		return 2
+	case errCobra:
+		return 3
 	case nil:
 		return 0
 	default:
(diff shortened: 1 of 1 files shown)


acbuild/acbuild.go
```diff
@@ -249,7 +249,10 @@ func main() {
 	// Make help just show the usage
 	cmdAcbuild.SetHelpTemplate(`{{.UsageString}}`)
 
-	cmdAcbuild.Execute()
+	err := cmdAcbuild.Execute()
+	if cmdExitCode == 0 && err != nil {
+		os.Exit(3)
```
jonboulle: why 3 and not 63 or 105?

dgonyeo: Because I know I used 1 in some places, and couldn't remember if I used 2. Why 63 or 105?

jonboulle: https://www.youtube.com/watch?v=MfB1VdNWtEU

acbuild/acbuild.go
```diff
@@ -249,7 +249,10 @@ func main() {
 	// Make help just show the usage
 	cmdAcbuild.SetHelpTemplate(`{{.UsageString}}`)
 
-	cmdAcbuild.Execute()
+	err := cmdAcbuild.Execute()
```
jonboulle: log err on debug?

dgonyeo: As far as I can tell errors here are directly from cobra (for example, if the called subcommand doesn't exist) and cobra will have already printed out an error and/or help page by this point.

dgonyeo:
I moved the 3 value into the `getErrorCode` function, so hopefully it's a little clearer that I'm just incrementing the returned exit codes whenever I come up with a different reason to not return 0.

jonboulle:
ok, lgtm

PR #180, opened 2016-02-22
Explain needed labels in getting started

EpocSquadron:
As written, the resulting image does not run. Rkt makes a hard requirement 
that an os and arch label are assigned to an image.


dgonyeo:
In the latest release of acbuild these labels should be automatically added for you when an empty build is started.

How did you run into a scenario where you were missing them?

EpocSquadron:
@dgonyeo I was running rkt 1.0.0 on Arch Linux, attempting to follow the steps in the guide. When I got it to build, rkt complained about the missing os and arch labels. I just tried again with rkt 1.1.0, which just hit the AUR today. I receive the same error.

acbuild.sh:
```bash
#!/bin/bash
set -e

acbuildend () {
    export EXIT=$?;
    acbuild --debug end && exit $EXIT;
}

acbuild --debug begin
trap acbuildend EXIT

# User entered acbuild commands go here

# Base this off of alpine for maximum hipness
acbuild set-name epocsquadron/ramp-dynamic
#acbuild label add os linux
#acbuild label add arch amd64
acbuild dependency add quay.io/coreos/alpine-sh

# Install all needed programs
acbuild run -- apk update
acbuild run -- apk add \
    apache2 \
    apache2-ssl \
    mysql \
    php \
    php-mysqli \
    php-gd \
    php-curl \
    php-apache2 \
    php-mcrypt \
    php-mysql \
    supervisor

# Make sure we can use .htaccess files
acbuild copy assets/apache/enable-htaccess.conf /etc/apache2/conf.d/enable-htaccess.conf

# Set the default timezone
acbuild copy assets/apache/timezone.ini /etc/php/conf.d/timezone.ini

# Default vhost
acbuild copy assets/apache/dynamic-vhost.conf /etc/apache2/conf.d/dynamic-vhost.conf
acbuild copy assets/apache/setDocRoot.php /etc/apache2/includes/setDocRoot.php

# Secure grants for MySQL
acbuild copy assets/mysql/grants.sql /etc/mysql/grants.sql

# Supervisor configuration
acbuild copy assets/supervisor/supervisor.conf /etc/supervisor.conf
acbuild copy assets/supervisor/mysql.conf /etc/supervisor/conf.d/mysql.conf
acbuild copy assets/supervisor/apache.conf /etc/supervisor/conf.d/apache.conf

# Set mount points
acbuild mount add www /var/www

# Expose ports
acbuild port add http tcp 80
acbuild port add https tcp 443

# Set init as supervisor
acbuild set-exec -- /usr/bin/supervisord -n -c /etc/supervisor/supervisord.conf

# Export the container
acbuild write ramp-dynamic.aci
```

rkt version:
```
rkt Version: 1.1.0
appc Version: 0.7.4
Go Version: go1.6
Go OS/Arch: linux/amd64
Features: -TPM
```


acbuild version:
```
acbuild version v0.1.1
appc version 0.7.1+git
```

dgonyeo:
Your issue is that acbuild is kinda old, the newest version is v0.2.2.
Where'd you get it from? Is there some AUR package I should bump the
version of?

EpocSquadron:
Ah. That makes sense. Indeed I never checked if the package was flagged out of date or not.

https://aur.archlinux.org/packages/acbuild/

Looks like someone has already put in a complaint but the maintainer hasn't taken action yet. The rkt package is much better maintained at this point (different maintainer). Not sure what action to take other than modify the PKGBUILD. I could publish it to the AUR but it would need a different name, which would not be preferable.

In any case, packaging for a specific distro isn't really your problem. Thanks for taking the time to help me figure this out.

Issue #156, opened 2015-12-20
begin: starting from tar doesn't create correct directory structure


ericchiang:
```console
$ mkdir foo
$ acbuild begin ./foo
$ ls .acbuild/currentaci/
manifest  rootfs
$ acbuild end
$ tar -cf foo.tar foo
$ acbuild begin ./foo.tar
$ ls .acbuild/currentaci/
foo
$ acbuild set-name foo
set-name: open .acbuild/currentaci/manifest: no such file or directory
```

ericchiang:
It looks like acbuild is treating `foo.tar` like an ACI.

To help identify these issues, can acbuild validate ACIs before unpacking them? 
("ERROR: foo.tar is not a valid ACI")

As for the archive stuff. It would be nice to be able to begin by using a tar as the rootfs. Could acbuild do one of the following?

* Determine if the file is an archive or an ACI based on file extension.
* Add a `--from-archive` flag.

Or I can just keep untaring my files manually and begin from a directory :)

dgonyeo:
Yup, acbuild doesn't currently support starting from a tarball of a rootfs. I'll add support for that, since you're not the only person who's recently run into this.

PR #181, opened 2016-02-24
begin: when provided file doesn't look like an ACI, it's a rootfs tar

dgonyeo:
This commit modifies begin such that after a local file has been given
to acbuild and it's been extracted, acbuild will check for the existence
of a manifest and a rootfs. If either is missing, it's assumed that the
tarball acbuild just extracted is the rootfs for the container, and it
will move the files into the rootfs directory and create an empty
manifest.

Fixes https://github.com/appc/acbuild/issues/156

### DIFF
--- a/lib/begin.go
+++ b/lib/begin.go
@@ -108,7 +108,45 @@ func (a *ACBuild) beginFromLocalImage(start string) error {
 	if finfo.IsDir() {
 		return fmt.Errorf("provided starting ACI is a directory: %s", start)
 	}
-	return util.ExtractImage(start, a.CurrentACIPath, nil)
+	err = util.ExtractImage(start, a.CurrentACIPath, nil)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range []struct {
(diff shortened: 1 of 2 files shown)


lib/begin.go
```diff
@@ -108,7 +108,45 @@ func (a *ACBuild) beginFromLocalImage(start string) error {
...
+		FileName string
+		FilePath string
+	}{
+		{"manifest file", path.Join(a.CurrentACIPath, aci.ManifestFile)},
+		{"rootfs directory", path.Join(a.CurrentACIPath, aci.RootfsDir)},
+	} {
+		_, err = os.Stat(file.FilePath)
+		switch {
+		case os.IsNotExist(err):
+			fmt.Fprintf(os.Stderr, "%s is missing, going to assume build is beginning with a tar of a rootfs\n", file.FileName)
```
jonboulle: nit: "%s missing, assuming build is beginning from" ...

jonboulle:
LGTM

PR #183, opened 2016-02-26
Update Vagrant script to use Rkt 1.0.0

urzds:
Currently the Vagrant install script uses Rkt 0.9.0. Version 1.0.0 was released recently, so update to that.


blalor:
1.1.0 was actually released yesterday. :-)

urzds:
> 1.1.0 was actually released yesterday. :-)

Submitted as #184, because renaming the branch confused GitHub.

Issue #98, opened 2015-11-08
Please "vendor" released/tagged dependencies whenever possible


onlyjob:
Downstream package maintainers should be able to reuse packages that are usually built from released/tagged versions instead of random snapshot.

Whenever possible please bundle (vendor) released/tagged versions of components in `Godeps`. Thanks.

jonboulle:
I don't really understand your question - we do vendor dependencies right now https://github.com/appc/acbuild/blob/master/Godeps/Godeps.json - no?

onlyjob:
Yes but bundled (vendored) version of _github.com/appc/spec_ is ahead of latest [release](https://github.com/appc/spec/releases). Presumably releases are suppose to be API-stable so vendoring random snapshot of "master" is not nice.

dgonyeo:
This is something I'll be interested in enforcing once acbuild comes a little farther along.

At this point in time a lot of its core features are still being created, which often requires patches to its dependencies. I'd prefer to not have acbuild development blocking on new releases of the appc spec or rkt, but once things become more stable it should be reasonable to stick to tagged releases of dependencies.

onlyjob:
Precisely. In this particular case I've noticed that _acbuild v0.1.1_ uses/requires unreleased snapshot of _appc/spec_ which is not compatible with its latest release. IMHO for _specification_ is even more important to stick to formal release than for ordinary library... There is no need to slow down development but release process needs to ensure that latest (formal, tagged) release of _acbuild_ is based upon formal release of appc specification and other libraries.

dgonyeo:
acbuild should now be vendoring a tagged version of both rkt (v1.2.0) and the appc spec (v0.7.4), so I'm going to close this.

onlyjob:
Thanks. :)

onlyjob:
Perhaps it would be a good time to tag new release of _acbuild_.

dgonyeo:
There's a few more things I'd like to get wrapped up before doing so. I've got a milestone for it:
https://github.com/appc/acbuild/milestones/v0.3.0

PR #157, opened 2015-12-21
appc: bump version of spec to 0.7.4

dgonyeo:
Addresses at least part of: https://github.com/appc/acbuild/issues/98

Would probably want to use a tagged release of rkt also to consider the issue fully addressed.

PR #186, opened 2016-02-26
*: replaced godep with glide

dgonyeo:
Godep was continually being a drain on development resources of acbuild,
and thus this commit replaces godep with glide. Glide takes advantage of
the golang vendor experiment, and thus this change will drop support for
versions of Go less than 1.5.

The biggest impact this change has is that dependencies are no longer
vendored with acbuild, and are instead fetched at compile time by the
build script. This is done by checking if the `vendor` directory exists,
and running `glide install` if it is not.

Fixes https://github.com/appc/acbuild/issues/98. Replaces https://github.com/appc/acbuild/pull/157.


jonboulle:
Why do you not want to vendor the deps?

Derek Gonyeo <notifications@github.com> schrieb am Fr., 26. Feb. 2016 20:23:

dgonyeo:
It's not something that's as nicely supported with glide. It leaves the .git directories in what it fetches, and I didn't see an easy way to disable that.

I see two arguments to not want to move away from vendoring our dependencies:
- What if one of our dependencies vanishes from github?
 -   We have a local copy on our machines (because it'll be sitting in acbuild's vendor directory), and can throw up a fork of the project.
- What if github goes down?
 -   This will only impact users who haven't yet built acbuild once, and chances are if someone who wants to build acbuild is at this stage then they don't have acbuild either, and won't be able to get it anyway.

I also think that glide's model will make it a good amount easier to work on both acbuild and one of acbuild's dependencies concurrently, since you can just `cd` into `vendor/github.com/the/dependency`, make changes, see the impact on acbuild, and when you're done commit and push inside the vendored repo.

sgotti:
Hi, as I also started experimenting with glide and I'd like to share what I learned since I'd prefer to vendor dependencies:

* Glide doesn't remove .git and other VCS directories. But you can easily do this and now there's a tool to do this:

http://engineeredweb.com/blog/2016/glide-strip-vendor-vcs/

* Glide doesn't remove unused packages so the created vendor, also with VCS data stripped out, remains with unneeded files (for example importing just the etcd client package imports the whole etcd project). The explanation is related to licensing and project modifications:

http://engineeredweb.com/blog/2016/go-why-not-strip-unused-pkgs/

But if you are sure of not causing license issues you can probably write a tool (I'm going to do this if it doesn't exists) that, given the glide.lock will remove all the files outside the import or subpackages (if the import have sub packages).

* Just for completeness (but it's a general info and not related to glide) as the glide faq says, vendoring is good for the final application but with a package meant to be used by other projects it can cause problems in some cases:
https://github.com/mattfarina/golang-broken-vendor

But this is a problem present also with old style Godep vendoring/rewriting.

jonboulle:
@sgotti thanks for letting us know your experience! Would you mind sharing
more about why you wanted to move away from Godep and how glide improves?

sgotti:
@jonboulle sure. I'm currently not in the position to suggest e tool over another. I was intrigued on how it tries to manage dependencies like npm, pip, bundler using a config file and lock files, it tries to implements semver dependencies and read config from other vendor tools like godeps, gb etc...

Obviously, since golang is really distributed and doesn't have a central repository, there are some difference with other languages/tools.

One interesting thing is that it doesn't needs packages in your gopath but it will check them out directly inside your project (or copy them from a cache dir or from the gopath but then checking out the right version inside you project).

On the other side, glide doesn't enforce you to commit your vendor directory. This is causing me some headache since many projects (myself too) wants to commit it (for a lot of good reasons), but having or not having nested vendor directories have a different compilation behavior. I opened github.com/Masterminds#303 to talk about this.

Additionally, if I want to commit, and I'm sure to not cause licenses violations, I'd like to remove as much as possible from the vendor directory. I wrote a tool to do this (https://github.com/sgotti/glide-vc) (with some changes in flight, see the opened PRs)

sgotti:
After triggering some discussions on Masterminds/glide#303, glide 0.10 has much improved the vendor handling (http://engineeredweb.com/blog/2016/glide-0.10.0/)

Now you can easily remove nested vendor folders. This obviously means that you won't have problems like [this](https://github.com/mattfarina/golang-broken-vendor) but you can't have multiple versions of the same package (since all deps are flattened). So if you have two dependencies that depends on incompatible version of `pkgA` you have to find a fix for making them work together.

actually my preferred workflow is this one:

* create a glide.yaml (`glide create`, if already using godeps it'll convert them from godeps)
* Update it (I prefer to specificy semvers tagged releases when possibile)
* `rm -rf vendor` (I like to treat my vendor as a scratch and be able to recreate it every time)
* `glide update -u -s -v --cache` # This will populate the vendor dir, remove VCS data and strip vendor (also undo godeps import rewriting). `--cache` is very useful to avoid git cloning deps every time I run it, you can also use `--cache-gopath` and `--use-gopath` instead of having a dedicate cache dir.
* Test, fix glide.yaml for incompatibilities (for example force subdependency versions if needed)
* Before committing `glide.yaml`, `glide.lock` and `vendor/`, super strip `vendor/` removing (if licenses will permit it) all unneeded files using glide-vc (https://github.com/sgotti/glide-vc):
`glide-vc --only-go --no-tests`

jonboulle:
cross-linking https://github.com/sorintlab/stolon/pull/131

dgonyeo:
Revisited this, tests now pass. Now vendors dependencies, unlike when I originally made this PR (uses glide flags to trip VCS info). Added `Documentation/development-guide.md` with a short guide on using glide.

dgonyeo:
Oh, and due to switching to using `vendor`, this commit drops go 1.4 from the travis tests.

jonboulle:
@dgonyeo there is a tonne of unused stuff in there - is this post-`glide-vc`?

dgonyeo:
No it's not. Just ran `glide-vc`, updated the docs to state that it should be used and how to use it, and added these changes to the PR.

PR #193, opened 2016-03-22
godep: bump the version of rkt to v1.2.0

dgonyeo:
Until a consensus is reached on [moving to glide](https://github.com/coreos/docs/issues/775), I'd like to bump the appc version (which required bumping the rkt version) used by acbuild. This allows for apps with an empty exec statement.


jonboulle:
Build broke

dgonyeo:
Fixed the build. Apparently I need `godep go build` now instead of `go build`.

jonboulle:
Eurgh. Instead, set GOPATH: https://github.com/appc/spec/pull/578/files#diff-b0da275520918e23dd615e2a747528f1L13

jonboulle:
Also if you're going to unrewrite paths too (which I think is probably a good idea at this stage), please add something to the commit message to that effect (or split it into two commits)

dgonyeo:
Tweaked the commit message, and decided to mess with the build/test scripts while I'm here.

jonboulle:
Can't actually review the important changes, thanks github :-(

dgonyeo:
The big change in the build/test scripts is this:
https://github.com/appc/acbuild/blob/1beb799d4e93c4ee9860156e03405cf7b2c49357/build#L6

jonboulle:
if it works, ship it

Issue #190, opened 2016-03-06
wrong arch label generated on arm(v7l)


godart:
Running `actool begin` on a Cubietruck ARMv7 computer fails with:
```
begin: json: error calling MarshalJSON for type types.Labels: bad arch "arm" for linux (must be one of: [amd64 i386 aarch64 aarch64_be armv6l armv7l armv7b])
```
It looks like the label value is generated from GOARCH, but not the architecture name returned by `uname -m`.

dgonyeo:
I guess the correct thing to do here is to add a case to check which arm platform acbuild is running on when GOARCH is set to "arm". I'll see if I can make a PR to do that.

PR #195, opened 2016-04-01
begin/copy: scan for arm executables, update arch label when found

dgonyeo:
The runtime.GOARCH value isn't accepted by the AppC spec as a value for
the arch label in a manifest when on an arm platform. This commit adds a
case to use "armv7l" whenever runtime.GOARCH is "arm". Additionally
(when on arm) this change scans all files at the build start and any
files copied in later for executables, and will switch the label between
"armv7l" and "armv7b" when it finds a little or big endian binary.

Fixes https://github.com/appc/acbuild/issues/190.

### DIFF
--- a/lib/begin.go
+++ b/lib/begin.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"unsafe"
 
 	"github.com/appc/acbuild/registry"
 	"github.com/appc/acbuild/util"
@@ -196,6 +197,22 @@ func (a *ACBuild) writeEmptyManifest() error {
 		return err
 	}
 
+	archvalue := runtime.GOARCH
+	if runtime.GOOS == "linux" && (archvalue == "arm" || archvalue == "arm64") {
+		var x uint32 = 0x01020304
+		test := *(*byte)(unsafe.Pointer(&x))
+		switch {
+		case test == 0x01 && archvalue == "arm":
+			archvalue = "armv7b"
+		case test == 0x04 && archvalue == "arm":
+			archvalue = "armv7l"
+		case test == 0x01 && archvalue == "arm64":
+			archvalue = "aarch64_be"
+		case test == 0x04 && archvalue == "arm64":
+			archvalue = "aarch64"
+		}
+	}
+
 	manifest := &schema.ImageManifest{
 		ACKind:    schema.ImageManifestKind,
 		ACVersion: schema.AppContainerVersion,
@@ -203,7 +220,7 @@ func (a *ACBuild) writeEmptyManifest() error {
 		Labels: types.Labels{
 			types.Label{
 				*archlabel,
-				runtime.GOARCH,
+				archvalue,
 			},
 			types.Label{
 				*oslabel,
--- a/Documentation/subcommands/label.md
+++ b/Documentation/subcommands/label.md
@@ -31,6 +31,22 @@ When an empty ACI is created with `acbuild begin`, by default the `os` and
 `arch` labels are created for you. Their default values are the current
 system's OS and architecture, as determined by golang's `runtime` package.
 
+## The Arch Label on ARM
+
+When `acbuild begin` is used without any starting image, `runtime.GOARCH` is
+used to determine the default value for the `arch` label. This doesn't
+differentiate between different arm versions, and according to the AppC spec
+acbuild needs to pick one of `armv6l`, `armv7l`, and `armv7b`.
+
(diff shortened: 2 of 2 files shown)


dgonyeo:
@godart: would you be willing to try out this modified version of acbuild, and see if this fixes your issue?

godart:
Thanks @dgonyeo. It works for me now.

I was trying myself to use the golang.org/x/sys/unix.Uname() function to determine the current architecture, but it did not compile due to bugs in golang.

anguslees:
Hrm - is it important that this works for cross-compile situations?  If we can assume we're running natively, then it should be easy to work out current endianness and even armv6/armv7 without walking the binaries - using logic like [this xgetgoarm() function](https://github.com/golang/go/blob/master/src/cmd/dist/util.go#L516)

I feel like at some point we just have to add this as a command line option - particularly when cross-compiling - and at that point it doesn't seem worthwhile to try to be as clever as you're being here.  I can't see how we can assert generations of arm architecture other than by just trusting the user to tell us how they compiled the binaries contained in this rootfs :/  (other than completely disassembling all the code we find, looking for armv7 constructs, and assuming we found it all reliably..)

TL;DR: Can we just have a command line option, and default to something based on querying the local CPU (via /proc/cpuinfo or similar)?

lib/begin.go
```diff
@@ -188,6 +189,22 @@ func (a *ACBuild) writeEmptyManifest() error {
 		return err
 	}
 
+	archvalue := runtime.GOARCH
+	if runtime.GOOS == "linux" && archvalue == "arm" {
+		archvalue = "armv7l"
+		walker := func(path string, info os.FileInfo, err error) error {
```
anguslees:
```go
var x uint32 = 0x01020304
switch *(*byte)(unsafe.Pointer(&x)) {
case 0x01:
    archvalue = "armv7b"
case 0x04:
    archvalue = "armv7l"
}
```

dgonyeo:
No matter what arch label acbuild tries to specify, the user can always override it with a manual `acbuild label add arch <value>`, so I think acbuild should work fine in the case of cross-compilation.

lib/begin.go
dgonyeo: What's this doing? Looking at the acbuild binary that's in memory for some specific value?

anguslees: Oh, it's just the usual way to detect local endianness - put a known multibyte value into memory, and then see how the compiler decided to actually store it.  In this case it's seeing if 0x01020304 was stored with 0x01 or 0x04 first.

anguslees: Another (better?) option might be to just use `uname -m` (aka `syscall.Uname` in golang).

s-urbaniak: We used the same trick in mesos-dns: https://github.com/mesosphere/mesos-dns/blob/883e8ace58008d18f35bca878175212eea5b6b1f/detect/masters.go#L107

Maybe we could factor this in a library function returning `binary.ByteOrder`?

dgonyeo:
Just changed this to be more sane, now it'll just look at the endinanness of `/proc/self/exe` (based on the elf data) and pick an armv7 label. Users wishing for an armv6 label can just manually set it.

@godart would you be willing to try out this new set of changes?

lib/begin.go
```diff
@@ -196,14 +197,31 @@ func (a *ACBuild) writeEmptyManifest() error {
 		return err
 	}
 
+	archvalue := runtime.GOARCH
+	if runtime.GOOS == "linux" && archvalue == "arm" {
+		elfFile, err := elf.Open("/proc/self/exe")
```
anguslees: .. so why do this rather than my 0x01020304 approach suggested [earlier](https://github.com/appc/acbuild/pull/195#discussion_r58650830)?

dgonyeo: Mostly because this was already written :p. I'll switch it over to use your method tomorrow morning.

dgonyeo:
Just pushed the change to use @anguslees's method for determining endianness.

anguslees:
lgtm (fwiw)

dgonyeo:
It would be real convenient if I had a couple of arm devices to test this on :/

jonboulle:
did you fill out the lab survey? ;-)

dgonyeo:
Just pushed changes to also handle aarch64 and aarch64_be, and tested it out on an aarch64 machine courtesy of redbeard. Looks like it works fine at least in that case.

jonboulle:
LGTM

godart:
Sorry. I've been away from my account for a while.
Works well for my Cubietruck (armv7l) as well.

PR #196, opened 2016-04-02
Add apt-get update to ubuntu

meganuke19:
After install ubuntu i could not install git with error "Package git is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or is only available from another source"


dgonyeo:
Cool, thanks!

PR #198, opened 2016-04-04
docs: add basic docs for isolator, mention capabilities

dgonyeo:
A common question is how to retain linux capabilities in an ACI via
acbuild, which is done with the isolator command.

This commit adds docs on the command, which includes a simple example
for retaining a capability.


jonboulle:
LGTM

Issue #187, opened 2016-03-02
vagrant: switch to using a wily image


dgonyeo:
Apparently ubuntu/vivid64 doesn't exist in some cases. Should switch to using a wily image.

https://github.com/coreos/rkt/issues/2243

saintx:
$ vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Box 'ubuntu/vivid64' could not be found. Attempting to find and install...
    default: Box Provider: virtualbox
    default: Box Version: >= 0
The box 'ubuntu/vivid64' could not be found or
could not be accessed in the remote catalog. If this is a private
box on HashiCorp's Atlas, please verify you're logged in via
`vagrant login`. Also, please double-check the name. The expanded
URL and error message are shown below:

URL: ["https://atlas.hashicorp.com/ubuntu/vivid64"]
Error: The requested URL returned error: 404 Not Found

PR #203, opened 2016-04-27
vagrant: use xenial64 instead of vivid64

dgonyeo:
The vivid64 image is no longer available.

Fixes https://github.com/appc/acbuild/issues/187.


jonboulle:
LGTM

Issue #217, opened 2016-06-09
Missing /etc/machine-id breaks systemd >= 230


lucab:
Recent systemd (>= 230) expects `/etc/machine-id` to exist inside the container, otherwise it will refuse to start with:
```
+ acbuild run /bin/bash
Failed to read machine ID from ~/.acbuild/currentaci/rootfs/etc/machine-id: No such file or directory
```

For reference, the issue and a similar patch are being discussed for rkt: https://github.com/coreos/rkt/pull/2440

PR #220, opened 2016-06-13
engine/systemdnspawn: copy in the host's /etc/machine-id for >= 230

dgonyeo:
Starting in systemd 230, systemd-nspawn requires the file
/etc/machine-id to be populated in the container for some reason. There
does not appear to be a flag to disable this, and the commit message I
found for the systemd commit that I think causes this is rather vague as
to why this is necessary.

To work around this, this commit copies /etc/machine-id in from the host
and removes it after the command has run.

Fixes https://github.com/appc/acbuild/issues/217.

### DIFF
--- a/engine/systemdnspawn/systemdnspawn.go
+++ b/engine/systemdnspawn/systemdnspawn.go
@@ -47,6 +47,25 @@ func (e Engine) Run(command string, args []string, environment types.Environment
 		}
 		nspawncmd = append(nspawncmd, "--chdir", workingDir)
 	}
+	if systemdVersion >= 230 {
+		machineIdFile := path.Join(chroot, "/etc/machine-id")
+		_, err := os.Stat(machineIdFile)
+		switch {
+		case os.IsNotExist(err):
+			err := os.MkdirAll(path.Dir(machineIdFile), 0755)
+			if err != nil {
+				return err
+			}
+			f, err := os.Create(machineIdFile)
+			if err != nil {
+				return err
+			}
+			f.Close()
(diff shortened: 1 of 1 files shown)


iaguis:
I think that if you just create an empty file, a machine ID will be generated by nspawn.

jonboulle:
it requires a file to exist, but the file can be empty?! that seems dumb

dgonyeo:
Looks like it still works when I just make an empty file. This is absurd, but it appears to work, so whatever. Just tweaked the commit to stat the file, and create it if it's not there.

engine/systemdnspawn/systemdnspawn.go
```diff
@@ -47,6 +47,25 @@ func (e Engine) Run(command string, args []string, environment types.Environment
 		}
 		nspawncmd = append(nspawncmd, "--chdir", workingDir)
 	}
+	if systemdVersion >= 230 {
+		machineIdFile := path.Join(chroot, "/etc/machine-id")
+		_, err := os.Stat(machineIdFile)
```
jonboulle: Rather unatomic but I guess it's okay for this use case.

jonboulle:
Has anyone tried asking upstream if this bizarreness can be reverted?

lucab:
@jonboulle (kind-of) author rationale at https://github.com/coreos/rkt/issues/2416#issuecomment-209889417

engine/systemdnspawn/systemdnspawn.go
dgonyeo: In this case there can only be one acbuild acting at a time on this build due to the lock it grabs when a command starts.

PR #222, opened 2016-06-15
*: store persists between builds, added commands to interact with store

dgonyeo:
This commit modifies acbuild to retain the images it fetches between
builds. These images are kept in `/var/lib/acbuild`. This results in a
better UX, as iterating on an acbuild script takes much less time when
the dependencies are not constantly re-downloaded.

The commands added to interact with the store are:
- **acbuild store fetch IMAGE**: fetches an image into the store
- **acbuild store rm IMAGE**: removes an image from the store
- **acbuild store list**: lists all images in the store, in the form of
  NAME:VERSION
- **acbuild store clear**: deletes all images in the store

### DIFF
--- a/acbuild/acbuild.go
+++ b/acbuild/acbuild.go
@@ -68,6 +68,7 @@ var (
 	contextpath    string
 	aciToModify    string
 	disableHistory bool
+	storeDir       string
 
 	cmdExitCode int
 
@@ -96,12 +97,13 @@ func init() {
 	cmdAcbuild.PersistentFlags().StringVar(&contextpath, "work-path", ".", "Path to place working files in")
 	cmdAcbuild.PersistentFlags().StringVar(&aciToModify, "modify", "", "Path to an ACI to modify (ignores build context)")
 	cmdAcbuild.PersistentFlags().BoolVar(&disableHistory, "no-history", false, "Don't add annotations with the command that was run")
+	cmdAcbuild.PersistentFlags().StringVar(&storeDir, "store", "", "Path to use for acbuild's store (defaults to /var/lib/acbuild)")
 
 	cobra.EnablePrefixMatching = true
 }
 
 func newACBuild() *lib.ACBuild {
-	return lib.NewACBuild(contextpath, debug)
+	return lib.NewACBuild(contextpath, storeDir, debug)
 }
 
 func getErrorCode(err error) int {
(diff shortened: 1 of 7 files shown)


jonboulle:
How can we prevent drift/duplication between this store and rkt's?

dgonyeo:
If by drift you mean versions of stored ACIs getting out of sync, perhaps `store list` could show either image IDs or import timestamps, and thus users can at least see when the versions are different?

And if by duplication you mean double storing ACIs on disk that are both in acbuild's store and rkt's store, I'm not too concerned by that. acbuild's store is just a cache of build dependencies, and can be safely deleted at any time without the user missing anything (as of this implementation, these all came from the network and could always just be fetched again when next needed).

jonboulle:
I meant more between having two kinds of store backends with similar and expanding features etc - especially if we're considering splitting it out of rkt as a shared component for other systems to use.

sgotti:
@jonboulle  +1. Since in my dreams this can also become (or be the base) for an oci-build tool :smile: I'll leverage future work using shared components like a fetcher, a store and a renderer (or something similar) (like proposed in https://groups.google.com/d/msgid/rkt-dev/3cea91ae-1c7f-4604-8cb8-db018822ad08%40googlegroups.com?utm_medium=email&utm_source=footer)

dgonyeo:
If rkt's store logic is going to be split out, then I think I'd like to move forward with this and when the split happens I'll look into switching acbuild over to using the logic for rkt's store. In its current form acbuild's store is very simple, and I don't currently have plans to change it much. Having this will also enable things like [local dependencies](https://github.com/appc/acbuild/issues/227), which would make the whole picture around multi-image builds a lot cleaner.

In its current state this is a completely separate store from rkt's on disk, and shouldn't trample over it in any way.

sgotti:
@dgonyeo My 2 cents:

the current rkt store implements some logic to get a good store discovery behavior. For example:

* the store should be able to provide the best matching aci. If you have a persistent store you can end up with multiple aci providing the same name and labels. Currently the acbuild registry.GetACI returns the first matching aci in a dir walk order (alphabetical) and that can be ok with a temporary store but can give issues with a persistent one. The rkt store, to avoid this, saves the import time and choose the newest aci. Additionally there's the (not really well documented but used) latest behavior (so the rkt store had to adapt to it).

I tried to document these appc "issues" here: https://docs.google.com/document/d/1dd19t2bMdnRnpdP-eYLzbhPO1NYYP1M1Xo6YQfhAxRw/edit and I'd like to hear your thoughts on it.
 
That's one of the reason why I'll strongly suggest to use the rkt store also for acbuild if this is becoming a persistent store.

biax:
Hoping to see local dependencies possible in the next acbuild version.

dgonyeo:
It's fairly high on my todo list, I hope to circle back around to another attempt at this soon.

sanmai-NL:
@dgonyeo: Using `rkt` store for this functionality appeals to me due to DRY.

dgonyeo:
@sanmai-NL yup, it's just more work on my part. This is on hold until acbuild has its OCI support merged in and rkt's store gets tweaked to support storing OCI images.

I'm going to close this PR and open a new one once those things happen.

Issue #86, opened 2015-10-30
*: recipe for running acbuild in container


jonboulle:
We should have acbuild packaged in an ACI and then be able to invoke it using any appc runtime, e.g. rkt. I expect this would just mount in the user's project/asset directly to a known location (`/data` or whatever) and then run the script they pass to it.

For example with rkt this might look like:
```
$ ls
app.js
build-nodejs.sh
$ rkt run --volume data,kind=host,source=$(pwd) appc.io/acbuild ./build-nodejs.sh
```

Where "data" would be defined as a mountpoint in the `appc.io/acbuild` image.
Then this would output the ACI in the same directory.

dgonyeo:
In a fedora 22 container:
```
[root@rkt-a29fb369-a1dd-498d-863f-5de4eb5fe7f7 /]# systemd-nspawn
Not running on a systemd system.
```

Perhaps once alternate execution environments is implemented this will be possible, but it doesn't look like this is feasible with the current state of acbuild.

jonboulle:
@dgonyeo how about a container based on CoreOS or something to pick up a more modern systemd-nspawn? That check was removed a while ago: https://github.com/systemd/systemd/commit/4f923a1984476de3441922ee5bf7102ebdd250ef

dgonyeo:
I made a container out of `coreos_developer_container.bin.bz2` from http://alpha.release.core-os.net/amd64-usr/current/, but when I attempt to run the nginx example in it I get the following:
```
root@rkt-98235c83-868a-48a4-8d17-8ce7dd5da102 ~ # acbuild begin
root@rkt-98235c83-868a-48a4-8d17-8ce7dd5da102 ~ # acbuild dep add quay.io/coreos/alpine-sh
root@rkt-98235c83-868a-48a4-8d17-8ce7dd5da102 ~ # acbuild run -- apk update
Downloading quay.io/coreos/alpine-sh: [========================] 2.65 MB/2.65 MB
Failed to open system bus: No such file or directory
Failed to create directory /root/.acbuild/target/sys/fs/selinux: Read-only file system
Failed to create directory /root/.acbuild/target/sys/fs/selinux: Read-only file system
/etc/localtime is not a symlink, not updating container timezone.
Failed to copy /etc/resolv.conf to /root/.acbuild/target/etc/resolv.conf: No such device or address
Attempted to remove disk file system, and we can't allow that.
run: exit status 1
```
Any clue what's going on here? I can't figure out why it would be a read-only file system.

[...]

PR #223, opened 2016-06-16
*: added a script mode

dgonyeo:
This commit adds a script mode to acbuild, whereby build scripts can be
written in an acbuild DSL and then run. When acbuild is run in this way,
acbuild will exec itself with the debug flag, a work-path flag, and the
contents of each line. This lives under the "script" subcommand, but
this subcommand will be implied if the run acbuild has the name
"acbuild-script" (this was necessary to support /usr/bin/env shebangs).
The apache example has been rewritten to be in the acbuild DSL.

With a DSL mode various enhancements are now possible (like per-command
build caches), but the immediate benefit is the fact that acbuild can
enforce cleanup of builds that use the script mode.


jonboulle:
Curious to get a review from @lucab on this

acbuild/acbuild.go
```diff
@@ -239,6 +244,12 @@ func runWrapper(cf func(cmd *cobra.Command, args []string) (exit int)) func(cmd
 }
 
 func main() {
+	if path.Base(os.Args[0]) == "acbuild-script" {
+		os.Args[0] = path.Join(path.Dir(os.Args[0]), "acbuild")
+		args := os.Args[1:]
+		os.Args = []string{os.Args[0], "script"}
+		os.Args = append(os.Args, args...)
+	}
```
lucab: I think you are kind of fighting with multicall here. I was in need of something similar so I just split https://github.com/coreos/rkt/pull/2815 of my working branch. Do you think it could make things easier here?

acbuild/script.go
```diff
@@ -0,0 +1,168 @@
...
+		stderr("script: %v", err)
+		return getErrorCode(err)
+	}
+	return 0
+}
+
+func execScript(rawScript []byte) error {
+	script := strings.Split(string(rawScript), "\n")
+	for i := 0; i < len(script); i++ {
+		script[i] = trimComments(script[i])
```
lucab: I guess you can skip empty and newline-only lines here.

acbuild/script.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	for _, line := range script {
```
lucab: Can this second line-by-line loop be squeezed into the first one above?

acbuild/script.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	tmpDir, err := ioutil.TempDir("", "acbuild")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpDir)
+	contextpath = tmpDir
+
+	for _, line := range script {
```
lucab: This is another line-by-line loop. From a quick glance it looks like all these three can be merged into a single one, or am I mislooking something?

acbuild/script.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func execACBuild(workPath, line string) error {
+	args := []string{"--debug", "--work-path=" + workPath}
+	args = append(args, tokenizeLine(line)...)
```
lucab: I think it may be worth having some further validation here and check that argv[1] is a valid subcommand. I think you may also want to check for the "script in script" (is it a valid usecase?).

acbuild/acbuild.go
sgotti: Perhaps I'm missing something, but looks like this is related to cobra/pflags argument parsing (so cobra will execute `cmdScript`) than multicall.

lucab:
I'm typically not a fan of DSL, but I have to say that I'd really like having this scripting mode in acbuild! I've left some comments inline related to multicall usage, input validation and reducing runtime complexity.

acbuild/acbuild.go
lucab: Yes, it is re-adjusting `os.Args` for cobra, but I think this is basically working-around/re-implementing multicall behavior.

sgotti: @lucab ok, so you're suggesting to register the cmdScript as a multicall with name `acbuild-script`. Now I understand coreos/rkt#2815 :smile:

lucab: (Sorry for having been cryptic) Yes! It probably won't solve the issue here (argv[0] adjusting for later re-exec) but I think it will help in better integrating with multicall machinery.

acbuild/script.go
dgonyeo: There's a check later on (right before it reexec's acbuild) to skip lines that are empty strings, and the `strings.TrimSpace` call at the end of `trimComments` should strip all whitespace, newlines or otherwise.

acbuild/script.go
dgonyeo: I'll merge the first loop, but I want this to be separate from the other loop so I can search the whole input for run commands before execing anything. Otherwise I could get a few commands deep before discovering you're trying to use run as non-root.

dgonyeo:
I think this implementation of a DSL is a fairly sane way to add one, since the semantics are identical to whatever shell script you'd otherwise write, and if you want more power or control over a build you can still write a shell script just fine.

acbuild/script.go
dgonyeo: I do think that the "script in script" use case is valid. If a user wishes to separate out portions of their build process into separate files, it's trivial enough to support here. However having said that, I hadn't considered that this was possible and the current implementation resulted in the nested script both creating a new build directory and calling `acbuild end` when it completed, so I just pushed some changes that results in acbuild setting an environment variable and then looking for this variable when running a nested script.

I also don't want to do preemptive validation at this point, as some commands have aliases and all commands can be selected with a shorter version of their name (as in "dep" is a valid way to refer to the "dependency" command). Due to this, I think it's easier to just exec acbuild and let cobra do the checking.

dgonyeo:
Ready for another review pass

acbuild/script.go
jonboulle: for i, s := range script

acbuild/script.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func runScript(cmd *cobra.Command, args []string) (exit int) {
+	if len(args) != 1 {
+		cmd.Usage()
+		return 1
+	}
+
+	scriptName := args[0]
+	rawScript, err := ioutil.ReadFile(args[0])
```
jonboulle: scriptName

acbuild/script.go
```diff
@@ -0,0 +1,188 @@
...
+		script[i] = trimComments(script[i])
+
+		if strings.HasPrefix(script[i], "run") && os.Geteuid() != 0 {
+			return fmt.Errorf("scripts using the run subcommand must be run as root")
+		}
+	}
+
+	var tmpDir string
+	nestedScript := false
+	if inheritedPath := os.ExpandEnv("$" + nestedScriptEnvVar); inheritedPath != "" {
```
jonboulle: look down two functions in the godoc ;-) - https://golang.org/pkg/os/#Getenv

acbuild/script.go
```diff
@@ -0,0 +1,188 @@
...
+		fmt.Printf("nested script! %s\n", inheritedPath)
```
jonboulle: !

acbuild/script.go
```diff
@@ -0,0 +1,188 @@
...
+		tmpDir, err = ioutil.TempDir("", "acbuild")
+		if err != nil {
+			return err
+		}
+		defer os.RemoveAll(tmpDir)
+		contextpath = tmpDir
+	}
+
+	for _, line := range script {
+		if line != "" {
```
jonboulle: let's try remove some of this indenting?
if line == "" {
  continue
}

acbuild/script.go
```diff
@@ -0,0 +1,188 @@
...
+			break
+		}
+	}
+	if trimAt != -1 {
+		line = line[:trimAt]
+	}
+	return strings.TrimSpace(line)
+}
+
+func tokenizeLine(line string) []string {
```
jonboulle: I'd like to see some nice table testing for this method. Tokenizer bugs are annoying.

acbuild/script.go
```diff
@@ -0,0 +1,188 @@
...
+	var tokens []string
+	var currentToken []rune
```
jonboulle: nit but bytes.Buffer (with WriteRune) would be more appropriate

acbuild/script.go
```diff
@@ -0,0 +1,188 @@
...
+	inSingleQuoteBlock := false
+	inDoubleQuoteBlock := false
+	isEscaped := false
+	for _, char := range line {
+		if isEscaped {
+			currentToken = append(currentToken, char)
```
jonboulle: What if someone tries to split a command across multiple lines?

jonboulle: Some cases to throw at you:

```
# split command
run my-command arg1 \
     arg2
```

```
# oops, forgot a quote
add-anno my-anno 'some val
```

```
# oops, superfluous quote
label foo bar"
```

acbuild/script.go
dgonyeo: I had forgotten that go lets you modify slices as you iterate over them. I'll change this.

dgonyeo:
Addressed nits, added the ability to have multi-line commands with trailing `\` characters, and added a suite of test cases to try to hit the edge cases in the tokenizer and related functions.

jonboulle:
@squeed another one it might be nice to have you review

acbuild/script_test.go
```diff
@@ -0,0 +1,251 @@
...
+		if err != c.err {
+			t.Errorf("error, expected:%v actual:%v", c.err, err)
+		}
+		if !equal(output, c.output) {
+			t.Errorf("output, expected:%v actual:%v", c.output, output)
+		}
+	}
+}
+
+// no really guys, this language is _great_
```
lucab: :smile:

lucab:
No further concerns from my side.

acbuild/script_test.go
iaguis: People always ask me, what about languages? I very quicky answer, we have the best languages, they're outstanding.

squeed:
This testcase fails for `TestTrimComments`, should it?

```
testcase{
    "this is a \"comment in a #string\".",
    "this is a \"comment in a #string\".",
},
```

squeed:
It might be worth the extra effort to throw together a yacc grammar for this...

dgonyeo:
No, that shouldn't fail. I had forgotten about comments in strings. Since my comment parsing logic was run before the logic that's checking for strings, I removed `trimComments`, added its functionality to `tokenizeLine`, and added the cases `TestTrimComments` was checking for to `TestTokenizeLine`.

cdrage:
AWESOME work @dgonyeo ! I've been waiting for a functionality like this to be added. Just when I get back into looking at `acbuild` again in order to implement a DSL, I see this! Thanks for the awesome work!

I'd love to starting throwing out some examples for for acbuild scripts, one suggestion I have is the ability to use capitalization for each command.

Ex:
```
#!/usr/bin/env acbuild-script
BEGIN
SET-NAME example.com/apache

DEP add quay.io/coreos/alpine-sh


RUN -- apk update
RUN -- apk add apache2
RUN -- /bin/sh -c "echo 'ServerName localhost' >> /etc/apache2/httpd.conf"

PORT add http tcp 80

MOUNT add html /var/www/localhost/htdocs

SET-EXEC -- /bin/sh -c "chmod 755 / && /usr/sbin/httpd -D FOREGROUND"

WRITE --overwrite apache-latest-linux-amd64.aci
END
```

Also, is there the ability to remove the `--` for each command as it'll be implied in the scripting language?

dgonyeo:
Thanks :)

I have no issue with capitalizing the subcommand, acbuild can just make the first token of each line lowercase before running the command.

As far as removing the `--`, that exists due to the need to differentiate between flags to acbuild and flags to the command you're running. Right now, you could have a script with this line:

```
run --working-dir=/opt -- mycoolapp --some-option .
```

Without the `--`, cobra will try to parse `--some-option` as a flag to acbuild.

Maybe a way to remove it would be to have acbuild insert a `--` before the first argument that doesn't start with a `-`? This would mean that if you want to run a binary that starts with the `-` character it wouldn't work how you expect, but in that case you can just provide an absolute path to the binary.

cdrage:
@dgonyeo Thanks for replying!

Yeah, removing `--` would be _great_ as it'll simplify the experience when building an `.aci`. 

Although you are right. It'll create some large differentiation between the command line as well as the scripting language. 

Would the best way be to create documentation explaining it? Similar to other composable formats (I'm thinking kubernetes) as long as we have a decent amount of documentation flowing it'll be easier for users to switch to building via a DSL.

dgonyeo:
It is documented [here](https://github.com/appc/acbuild/blob/master/Documentation/subcommands/run.md#options-parsing), if you think the description there isn't sufficient I could always write a little more on the reasoning behind it.

And I'm comfortable with UX related diverges from the cli syntax. It should still be pretty clear the relation between the two syntaxes, and the `--` has been a common complaint.

dgonyeo:
Just pushed a commit that lowercases the first token of a line, and inserts a `--` before the first argument on a `run` command that's lacking a `--`.

squeed:
This seems reasonable to me.

A question: what are the security implications of running untrusted acbuild scripts?

dgonyeo:
Were I to hand someone a malicious acbuild script, a couple things come to mind:
- I could copy any files from your host and ship them off somewhere
- I could attempt to break out of systemd-nspawn or the chroot engine (the latter is probably easier), and if I did I'd have root on the box
- I could write an ACI to any path, potentially overwriting things you'd find important

So I'd highly recommend reviewing an acbuild script before running it. There's an issue from forever ago (here: https://github.com/appc/acbuild/issues/86) for running acbuild inside of rkt, and with the recent chroot engine this should now be much easier to implement, if someone wanted a sandboxed way to run an acbuild script.

squeed:
Ahh, all good points! Thanks!

dgonyeo:
Review ping?

jonboulle:
@squeed @lucab @alban

squeed:
👍 LGTM!

examples/apache/build-apache.acb
```diff
@@ -0,0 +1,28 @@
...
+begin
+
+# Name the ACI
+set-name example.com/apache
+
+# Based on alpine
+dep add quay.io/coreos/alpine-sh
+
+# Install apache
+run -- apk update
```
cdrage: doesn't `--` need to be removed (since you updated it to assume it when running `run` options, etc?)

cdrage: ahhhh, saw your comment that said it will automatically add if it's missing, nvm! so this should work regardless.

dgonyeo: Just double checked, this script runs fine.

lucab:
LGTM too.

PR #226, opened 2016-06-24
engine: added a chroot engine

dgonyeo:
Added a chroot run engine to acbuild. This will allow people who have
problems with systemd-nspawn and people who don't have systemd to use
acbuild run, by specifying `--engine=chroot` on any run commands.

### DIFF
--- a/engine/chroot/chroot-child.go
+++ b/engine/chroot/chroot-child.go
@@ -1,0 +1,89 @@
+// Copyright 2016 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package chroot
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
+	"syscall"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	cmdACBuildChroot.PersistentFlags().StringVar(&flagCmd, "cmd", "", "The command to run")
+	cmdACBuildChroot.PersistentFlags().StringSliceVar(&flagArgs, "args", nil, "arguments for the command")
+	cmdACBuildChroot.PersistentFlags().StringSliceVar(&flagEnv, "env", nil, "environment for the command")
+	cmdACBuildChroot.PersistentFlags().StringVar(&flagChroot, "chroot", "", "dir to chroot into")
+	cmdACBuildChroot.PersistentFlags().StringVar(&flagWorkingDir, "working-dir", "", "working directory for the command")
+}
+
+var (
+	flagCmd          string
+	flagArgs         []string
+	flagEnv          []string
+	flagChroot       string
+	flagWorkingDir   string
+	cmdACBuildChroot = &cobra.Command{
+		Use: "",
+		Run: runChroot,
+	}
+)
+
+func stderr(format string, a ...interface{}) {
+	out := fmt.Sprintf(format, a...)
+	fmt.Fprintln(os.Stderr, strings.TrimSuffix(out, "\n"))
+}
+
+func errAndExit(format string, a ...interface{}) {
+	stderr(format, a...)
+	os.Exit(1)
+}
+
+func runChroot(cmd *cobra.Command, args []string) {
+	runtime.LockOSThread()
+	err := syscall.Chroot(flagChroot)
+	if err != nil {
+		errAndExit("couldn't chroot: %v", err)
+	}
+	err = os.Chdir("/")
+	if err != nil {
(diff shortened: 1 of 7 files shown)


jonboulle:
I'm a little split between this approach and just relying on & execing `/bin/chroot`, since then we wouldn't need any multicall shennanigans..

dgonyeo:
I can switch it to using `/bin/chroot` if you want, but I'm not sure how the `--working-dir` command would work, since it looks like `/bin/chroot` doesn't let me set that.

engine/chroot/chroot-child.go
```diff
@@ -0,0 +1,89 @@
...
+	err := syscall.Chroot(flagChroot)
+	if err != nil {
+		errAndExit("couldn't chroot: %v", err)
+	}
+	err = os.Chdir("/")
+	if err != nil {
+		errAndExit("couldn't cd: %v", err)
+	}
+
+	//if flagWorkingDir != "" {
```
jonboulle: remove

jonboulle:
@squeed one more for you

engine/chroot/chroot-child.go
dgonyeo: Whoops, meant to uncomment that.

squeed:
Does the build script need to be updated for the new binary? When testing this, I get `run: exec: "acbuild-chroot": executable file not found in $PATH`

dgonyeo:
This requires an `acbuild-chroot` symlink (just back to acbuild) somewhere in your PATH, which is something I should probably document.

dgonyeo:
I just updated the build script to create a hard link for `acbuild-chroot`. Figured it would be less fragile than a symlink for when people attempt to package it.

squeed:
Oh, hey, another thought! When using `chroot` mode, you don't need to be root anymore! you just need `CAP_SYS_CHROOT`. Might be nice to remove all of the `os.Geteuid()` calls.

jonboulle:
https://github.com/coreos/rkt/blob/master/pkg/sys/capability.go#L25

squeed:
FWIW, DNS resolution in the `chroot` environment doesn't work, as there is no `/etc/resolv.conf`. Building a working environment is, of course, one of the advantages of `systemd-nspawn`, which creates this file along with lots of other setup.. This isn't a bug per se, but it does mean the engines are not equivalent. Of course, if the two engines were made to be equivalent, then it just means we went and re-implemented systemd, which seems sort of silly.

OTOH, if the `chroot` engine should be able to run fairly typical commands in the `acbuild` process, e.g. `apt-get update`, then this will have to be fixed. Deciding where to draw the line is not easy.

dgonyeo:
I think making the two engines equivalent is worthwhile, so I'd like to have the engine copy in stuff like that (and probably remove it later). systemd-nspawn has been a [constant](https://github.com/appc/acbuild/issues/140) [source](https://github.com/appc/acbuild/issues/131) of [bugs](https://github.com/systemd/systemd/issues/1400) in acbuild with [flags changing over time](https://github.com/appc/acbuild/blob/master/engine/systemdnspawn/systemdnspawn.go#L41) and [other such things](https://github.com/appc/acbuild/issues/217), so having an engine with no external dependencies would at least provide a workaround for the next user that reports `acbuild run` isn't working due to systemd-nspawn changing something again.

Also I think acbuild will still need root, due to its reliance on mounting overlayfs to keep layers separated during run commands, and iirc the only capability that lets you mount is CAP_SYS_ADMIN, which gives you so many things it's scary from a security perspective.

iaguis:
lgtm

dgonyeo:
Just tweaked the commit so it stats `/etc/resolv.conf` inside the container, and if it doesn't exist it'll copy it in from the host and remove it after the command has exited. Could I get another LGTM?

engine/chroot/chroot.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type Engine struct{}
+
+func init() {
+	multicall.Add("acbuild-chroot", cmdACBuildChroot.Execute)
+}
+
+func (e Engine) Run(command string, args []string, environment types.Environment, chroot, workingDir string) error {
+	resolvConfFile := path.Join(chroot, "/etc/resolv.conf")
```
iaguis: filepath

engine/chroot/chroot.go
```diff
@@ -0,0 +1,90 @@
...
+	_, err := os.Stat(resolvConfFile)
+	switch {
+	case os.IsNotExist(err):
+		err := os.MkdirAll(path.Dir(resolvConfFile), 0755)
```
iaguis: filepath

iaguis:
> it'll copy it in from the host and remove it after the command has exited

The nspawn engine leaves some files in `/etc`: the localtime link, `machine-id` and `resolv.conf`.

OTOH it seems cleaner to remove it when the command has exited.

iaguis:
a couple of nits but lgtm after that

dgonyeo:
Renamed the variables, and I'd prefer to have the engines clean up after themselves. Guess that means there might be a little more work to do for the systemd-nspawn one.

Can I get an LGTM?

engine/chroot/chroot.go
iaguis: I meant using `filepath.Join`, not changing var's name :D

dgonyeo: Oh wow that makes way more sense.

Fixed.

iaguis:
lgtm

Issue #235, opened 2016-07-16
BUG: acbuild ignores proxy


Andrei-Pozolotin:
`http.ProxyFromEnvironment`

see
https://github.com/appc/acbuild/blob/master/registry/fetch.go#L357

compared to
https://github.com/appc/spec/blob/master/discovery/http.go#L58

dgonyeo:
It looks like the `http.DefaultTransport` sets the proxy to be `ProxyFromEnvironment`, so I should just need to do that when the insecure flag is set and I don't use the default transport. I'll make a PR.

dgonyeo:
@Andrei-Pozolotin any chance you'd be willing to see if https://github.com/appc/acbuild/pull/239 fixes this for you?

Andrei-Pozolotin:
@dgonyeo 

1) thank you for the PR

2) test result: it does partially resolve the issue

3) remaining problem is: `Proxy: http.ProxyFromEnvironment` 
should be applied regardless of `--insecure` /  `r.Insecure`, in other words: "proxy always applies"
https://github.com/appc/acbuild/blob/master/registry/fetch.go#L363

Andrei-Pozolotin:
1) another issue came up, which has to do with `quay.io` servers, 

2) for example, a portion of build script:
```
acbuild dep add quay.io/coreos/alpine-sh
```

3) produces the following proxy log:
```
2016-Jul-18 20:47:16    211 192.168.1.103 TAG_NONE/200 0 CONNECT quay.io:443 - HIER_DIRECT/174.129.232.79 -
2016-Jul-18 20:47:16    123 192.168.1.103 TCP_MISS/302 719 GET https://quay.io/coreos/alpine-sh?ac-discovery=1 - HIER_DIRECT/174.129.232.79 text/html
2016-Jul-18 20:47:16     79 192.168.1.103 TCP_MISS/200 17961 GET https://quay.io/repository/coreos/alpine-sh?tag=latest - HIER_DIRECT/174.129.232.79 text/html
2016-Jul-18 20:47:16    144 192.168.1.103 TAG_NONE/200 0 CONNECT quay.io:443 - HIER_DIRECT/174.129.232.79 -
2016-Jul-18 20:47:16    134 192.168.1.103 TCP_MISS/302 1127 GET https://quay.io/c1/aci/quay.io/coreos/alpine-sh/latest/aci/linux/amd64/ - HIER_DIRECT/174.129.232.79 text/html
[...]

PR #239, opened 2016-07-18
registry: Fixed proxy use when insecure flag is set

dgonyeo:
Fixes https://github.com/appc/acbuild/issues/235. Hopefully.


iaguis:
LGTM.

Issue #238, opened 2016-07-18
BUG: Isolator commands are completely broken on master


monder:
## Bug ##

Isolator commands do not work.

### Expected Behavior ###
```
echo '{ "set": ["CAP_NET_BIND_SERVICE"] }' | acbuild --debug os/linux/capabilities-retain-set -
Adding isolator "os/linux/capabilities-retain-set"="{\"set\":[\"CAP_NET_BIND_SERVICE\"]}\n"
```
Manifest should contain the specified isolator.

### Actual Behavior ###
```
echo '{ "set": ["CAP_NET_BIND_SERVICE"] }' | acbuild --debug os/linux/capabilities-retain-set -
Adding isolator "os/linux/capabilities-retain-set"="{\"set\":[\"CAP_NET_BIND_SERVICE\"]}\n"
isolator add: json: error calling MarshalJSON for type *types.App: invalid isolator
```
Manifest does not contain any isolators.

### Other Information ###

After https://github.com/appc/acbuild/commit/993fbd1d349bb64b9096d80def4d4ff2bd3d2137 the [App](https://github.com/appc/acbuild/blob/993fbd1d349bb64b9096d80def4d4ff2bd3d2137/vendor/github.com/appc/spec/schema/types/app.go#L24) now tries to [validate isolators](https://github.com/appc/acbuild/commit/993fbd1d349bb64b9096d80def4d4ff2bd3d2137#diff-0e2735b8ddfbc0f8c863292d19e4df9aR89). Obviously the validation fails since we are [assigning isolator value](https://github.com/appc/acbuild/blob/993fbd1d349bb64b9096d80def4d4ff2bd3d2137/lib/isolator.go#L71) to [ValueRaw](https://github.com/appc/spec/blob/master/schema/types/isolator.go#L147) when its explicitly said [not to do so.](https://github.com/appc/spec/blob/master/schema/types/isolator.go#L143)

I don't understand what's the purpose of hiding the `value` in spec :(

PR #240, opened 2016-07-18
isolator: fixed isolator, added test

dgonyeo:
`acbuild isolator add ...` was broken, producing the error:

```
isolator add: json: error calling MarshalJSON for type *types.App: invalid isolator
```

This commit fixes this, and adds a test for isolator add to be able to
catch this type of regression in the future.

Fixes https://github.com/appc/acbuild/issues/238.

### DIFF
--- a/lib/isolator.go
+++ b/lib/isolator.go
@@ -16,6 +16,7 @@ package lib
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/appc/acbuild/util"
 
@@ -64,12 +65,27 @@ func (a *ACBuild) AddIsolator(name string, value []byte) (err error) {
 		if s.App == nil {
 			s.App = newManifestApp()
 		}
+		_, ok := types.ResourceIsolatorNames[*acid]
+		if !ok {
+			_, ok = types.LinuxIsolatorNames[*acid]
+			if !ok {
+				return fmt.Errorf("unknown isolator name: %s", name)
+			}
+		}
+		i := &types.Isolator{
+			Name:     *acid,
+			ValueRaw: &rawMsg,
+		}
+		blob, err := json.Marshal(i)
+		if err != nil {
+			return err
+		}
+		err = i.UnmarshalJSON(blob)
+		if err != nil {
+			return err
+		}
 		removeIsolatorFromMan(*acid)(s)
-		s.App.Isolators = append(s.App.Isolators,
-			types.Isolator{
-				Name:     *acid,
-				ValueRaw: &rawMsg,
-			})
+		s.App.Isolators = append(s.App.Isolators, *i)
 		return nil
 	}
 	return util.ModifyManifest(fn, a.CurrentACIPath)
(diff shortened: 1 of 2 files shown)


monder:
That certainly fixes the problem, but maybe we should change the spec a bit?

Current `spec` to my mind is kindof an anti-pattern.

For example there is a public field that shouldn't be public
Every object that implements the `Isolator` interface have to carry value hacks to [their](https://github.com/appc/spec/blob/de6f4f252da48aee7e6b60842ba916eea527eef3/schema/types/isolator_resources.go#L147) [implementation](https://github.com/appc/spec/blob/65a06aa441e4b1550f5c3c292b1f46c499277d00/schema/types/isolator_linux_specific.go#L278) and always to no forget it.
Maybe its time to refactor some of it before it gets out of control with all that Marshal/Unmarshal magic?

What I suggest is to change [this line](https://github.com/appc/spec/blob/d3cde0720bccadccc5ed60b8a70eb3dd3afb1d91/schema/types/isolator.go#L153) to be something like this: https://play.golang.org/p/7aOR3Bonc9 
That way the Marshal/Unmarshal logic will be kept privately in a single file as it should have been.

If its possible I can work on the POC PR.

dgonyeo:
I'd be in favor of applying a fix for this issue with the current spec, and then revising the spec to make this fix less of an anti-pattern. If you'd be willing to make a PR against the spec that would be awesome.

lib/isolator.go
```diff
@@ -64,12 +65,24 @@ func (a *ACBuild) AddIsolator(name string, value []byte) (err error) {
 		if s.App == nil {
 			s.App = newManifestApp()
 		}
+		_, ok := types.LinuxIsolatorNames[*acid]
```
monder: What about ResourceIsolatorNames?

dgonyeo: I'm not certain if the linux isolators make it into there. I just switched it over and now the test that I added with this PR fails with:

```
=== RUN   TestAddIsolator
--- FAIL: TestAddIsolator (0.02s)
	isolator_test.go:55: non-zero exit code of 1: stdout:
		
		stderr:
		isolator add: unknown isolator name: os/linux/capabilities-retain-set
		
	
```

monder: Right, the linux isolators are not there, the same reason as resource isolators are not in `LinuxIsolatorNames`.

```
$ git checkout origin/pr/240
$ ./build
Building acbuild...
$ ./bin/acbuild begin
$ echo '{ "set": ["CAP_IPC_LOCK"] }' | ./bin/acbuild isolator add "os/linux/capabilities-retain-set" -
$ echo '{ "limit": "20m" }' | ./bin/acbuild isolator add "resource/cpu" -
isolator add: unknown isolator name: resource/cpu
```

iaguis: Shouldn't we then check against both?

dgonyeo: Yeah. Guess I got confused :/

jonboulle:
LGTM

PR #241, opened 2016-07-18
util: unwrap errors from rkt/pkg/tar

cgonyeo:
errors are wrapped using the hashicorp errwrap package in
github.com/coreos/rkt/pkg/tar, which acbuild uses. This means that
whenever there is an error untarring something, the errors that users
report have no information about what actually went wrong. This commit
fixes that.


cgonyeo:
This is rather old and this project is now unmaintained, so closing.

PR #242, opened 2016-07-27
do not install git again if it exists

bcg62:
Currently the which git check is irrelevant. This combines the two actions in the same logic. 

$ true || echo "false" && echo "not needed"
not needed

$ true || { echo "false" && echo "not needed"; }
$


jonboulle:
lgtm, thanks for catching!

PR #251, opened 2016-09-12
enable ppc64le architecture

leitao:
This just add ppc64le support for acbuild. Currently, it does
not support ppc64le architecture, failing with the following error:

begin: json: error calling MarshalJSON for type types.Labels: bad arch "ppc64le" for linux (must be one of: [amd64 i386 aarch64 aarch64_be armv6l armv7l armv7b])

### DIFF
--- a/vendor/github.com/appc/spec/schema/types/labels.go
+++ b/vendor/github.com/appc/spec/schema/types/labels.go
@@ -21,7 +21,7 @@ import (
 )
 
 var ValidOSArch = map[string][]string{
-	"linux":   {"amd64", "i386", "aarch64", "aarch64_be", "armv6l", "armv7l", "armv7b"},
+	"linux":   {"amd64", "i386", "aarch64", "aarch64_be", "armv6l", "armv7l", "armv7b", "ppc64le"},
 	"freebsd": {"amd64", "i386", "arm"},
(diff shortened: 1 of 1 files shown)


jonboulle:
Thanks for the patch. What we really need to do here is bump the vendored appc spec package (to v0.8.7, which includes the ppc architectures), instead of editing it manually in place. Would you be up for trying this? The instructions are here:
https://github.com/containers/build/blob/master/Documentation/development-guide.md

leitao:
Hello Jonathan, I understand that the appc spec was already bumped to v0.8.7 and we still seeing this issue. Any further recommendation on how to get this fixed?

jonboulle:
I got an email notification for this but don't see a comment (was it deleted?)

leitao:
Hello @jonboulle, yes. I posted something and deleted.

You are correct, let's delete this feature request because the right way to solve it is bumping the appc spec version, as you recommended.

We still need add ppc64le on the rkt source code, adding 'ppc64le' at ValidOSArch (inside 'stage0/arch.go'), but probably @rmatinata will open a feature request for that.

Thank you,
Breno

jonboulle:
Thanks for the update - if you need any more guidance with the spec bump
please LMK

PR #254, opened 2016-09-14
Support HTTP Basic authentication when fetching dependencies

ragalie:
Hello!

I need to be able to supply HTTP Basic auth as part of the discovery process, so I modeled an implementation off of the `rkt` file format.

Can you take a look and let me know if this is along the lines of what you were thinking?

Thanks!

### DIFF
--- a/registry/auth-config.go
+++ b/registry/auth-config.go
@@ -1,0 +1,81 @@
+package registry
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+)
+
+type HostHeaders map[string]http.Header
+
+type AuthConfig []AuthFile
+
+type AuthFile struct {
+	Domains     []string    `json:"domains"`
+	Credentials Credentials `json:"credentials"`
+}
+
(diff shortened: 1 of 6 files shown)


ragalie:
@jonboulle is this something you'd consider including?

dgonyeo:
Wow, so sorry about not getting around to this. I'm just now getting back into acbuild development. I'll review this now.

dgonyeo:
Would you be willing to move the `authConfigDir` into the `ACBuild` struct in `lib/common.go`, instead of having it be an argument to `Begin`? I feel as if it belongs there better.

registry/auth-config.go
```diff
@@ -0,0 +1,81 @@
...
+	User     string `json:"user"`
+	Password string `json:"password"`
+}
+
+func ReadAuthConfig(directory string) (*AuthConfig, error) {
+	authConfig := AuthConfig{}
+
+	err := filepath.Walk(directory, func(path string, file os.FileInfo, err error) error {
+		if err != nil {
+			return nil
```
dgonyeo: Maybe it would be better to log here what went wrong?

PR #256, opened 2016-09-20
Add documentation for acbuild script

cdrage:


cdrage:
@dgonyeo Is it possible to add this to the `acbuild script --help` command too, or would it be too much? (since acbuild has no man pages)

dgonyeo:
Not sure, it would make it the largest man page.

Maybe including a link to the [GitHub documentation](https://github.com/containers/build/tree/master/Documentation) in the help page would help at the very least.

acbuild does also have the [ability to generate man pages](https://github.com/containers/build/blob/master/build-man-pages), but the information is the same as on the help pages.

cdrage:
@dgonyeo ah okay, yeah, adding a link to the github documentation would be a good half-way point for this.

cdrage:
@jonboulle quick review perhaps? :)

jonboulle:
lgtm, thanks

PR #257, opened 2016-09-25
Vagrant on OS X tweaks

alindeman:
Fixes a couple issues I ran into trying to build acbuild with Vagrant/VirtualBox on OS X.

* Hard links are not allowed on an OS X shared folder

  Fixes issue with building acbuild on Vagrant on OS X:
```
==> default: ln: failed to create hard link '/vagrant/bin/acbuild-script'
=> '/vagrant/bin/acbuild'
==> default: : Operation not permitted
```

* vagrant build uses go 1.5 (matches build-docker)
* Only attempt to add the group `rkt` if it doesn't currently exist (`vagrant provision` succeeds after initial provision)


jonboulle:
lgtm, thanks

PR #266, opened 2016-11-10
Update clone destination

androm3da:
The instructions don't work as-is.  It's a simple matter, presumably the repo was formerly called "acbuild".  I think it makes more sense to clone it into a dir called `acbuild` than to update the rest of the instructions.


dgonyeo:
Yup, you would be right. Thanks for fixing this!

PR #268, opened 2016-11-12
Fix relative links to docs

androm3da:
In " Specifics on using it are here," 'here' gets rendered as a link to https://github.com/containers/build/blob/master/Documentation/Documentation/subcommands/dependency.md


dgonyeo:
Whoops, yeah that's a problem. Thanks for the PR!

Issue #265, opened 2016-11-02
be able to set the supplementaryGIDs field in the manifest


dgonyeo:

GrantSeltzer:
@dgonyeo Thanks for the merge! This issue can be closed.

PR #269, opened 2016-11-17
Add command for setting supplementary GIDs

GrantSeltzer:
Handles #265

One thing that I questioned adding is having a check for [NGROUPS_MAX ](https://www.j3e.de/ngroups.html), interested to hear thoughts.

Signed-off-by: grantseltzer <grantseltzer@gmail.com>

### DIFF
--- a/acbuild/set-supplementary-groups.go
+++ b/acbuild/set-supplementary-groups.go
@@ -1,0 +1,65 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


acbuild/set-supplementary-groups.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if debug {
+		stderr("Setting supplementary groups to ", args)
+	}
+
+	var intArgs = []int{}
+	for _, stringArg := range args {
+		intArg, err := strconv.Atoi(stringArg)
+		if err != nil {
+			return getErrorCode(err)
```
dgonyeo: Can you add a `stderr` call right before this to explain to the user why we're exiting?

dgonyeo:
Thanks so much for the PR! One tiny nit, but aside from that LGTM.

acbuild/set-supplementary-groups.go
GrantSeltzer: Done.

GrantSeltzer:
@dgonyeo No problem, fixed the issue.

PR #271, opened 2016-11-21
Fix Begin to return error generated when starting from local dir

### LINKED ISSUE #270: Acbuild begin - Errors from beginFromLocalDirectory are being swallowed in Begin
Hi, 

I was trying to start a new build from a directory with `acbuild begin /path/to/rootfs` and it was only creating about half of the directories under `.acbuild/currentaci/rootfs` that had existed in the source directory.  After a little debugging, it looks like any errors that occur in beginFromLocalDirectory [https://github.com/containers/build/blob/master/lib/begin.go#L161](url) are swallowed in Begin [https://github.com/containers/build/blob/master/lib/begin.go#L96](url).  In this case it was swallowing a permission error when trying to copy the directory.

My Go-fu is pretty weak, but perhaps https://github.com/containers/build/blob/master/lib/begin.go#L96 should look like:
```
err = a.beginFromLocalDirectory(start)

if err != nil {
	return err
}
```

Thanks!

tjabell:
Fixed #270

### DIFF
--- a/lib/begin.go
+++ b/lib/begin.go
@@ -93,7 +93,7 @@ func (a *ACBuild) Begin(start string, insecure bool) (err error) {
 			case err != nil:
 				return err
 			case finfo.IsDir():
-				a.beginFromLocalDirectory(start)
+				return a.beginFromLocalDirectory(start)
 			default:
 				return a.beginFromLocalImage(start)
 			}


jonboulle:
lgtm, thanks so much!
@dgonyeo ?

dgonyeo:
Thanks so much for fixing this. LGTM.

PR #276, opened 2016-11-30
OCI support

dgonyeo:
This commit adds support to acbuild for building OCI compatible
container images. A new flag has been added to the `begin` command,
`--build-mode`, to specify when a build is started if it is for creating
an appc-compliant image or an OCI-compliant image.

The isolator, dependency, set-supp-groups, set-event-handler, and
set-name subcommands only work under the appc build mode.

All other subcommands work under both build modes, but obviously some
will have slightly different semantics.

A new subcommand named `layer` has been added. This subcommand only
works under oci build mode. It adds a new layer to the image, and future
file changes (via copy or run) will occur in the new layer.

A new flag, `--config`, has been added to the `cat-manifest` subcommand,
so that when the build mode is oci the config can be printed as well.


lib/appc/annotation.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2016 The appc Authors
```
jonboulle: why did this get renamed to the singular annotation but dependencies stayed plural? :stuck_out_tongue:

dgonyeo: Because more things are singular than plural, and something something consistency. I'll singular-ize dependencies too.

dgonyeo:
@jonboulle I just updated this PR, OCI support should be fully usable now and I think the changes are ready for review.

Things left for me to do:
- the `label` subcommand doesn't function under build mode oci
- the `--modify` flag won't work with oci images
- update documentation
- write OCI examples

vbatts:
Haven't done a line-by-line review, but tested it out and seems to be a great start.
Some of the cli usage needs to become more generic, i.e. where it references writing out an ACI may need to be "a container image" or similar?

vbatts:
LGTM

dgonyeo:
Yup, totally agree on the cli usage becoming more generic. When I go through and update all of the documentation to describe these changes, I'll make sure that includes the `--help` output for all the subcommands.

acbuild/cat-manifest.go
```diff
@@ -34,6 +33,7 @@ var (
 func init() {
 	cmdAcbuild.AddCommand(cmdCat)
 	cmdCat.Flags().BoolVar(&prettyPrint, "pretty-print", false, "Print the manifest with whitespace")
+	cmdCat.Flags().BoolVar(&printConfig, "config", false, "(oci only) Prints the image config instead of the manifest")
```
lucab: What about turning this into a `--type` or `--mode` enum? Boolean flags are harder to accomodate/extend for future cases.

dgonyeo: Changed it into `--file`, that accepts `"manifest"` or `"config"`.

dgonyeo:
Just finished going over all the docs updating them to be in line with these changes

blalor:
Will this result in a new release?

dgonyeo:
When this gets merged in, yes. One thing that I'm going to make a separate PR for is support for `begin`ing from remote OCI images, but if it's looking like I won't be able to add support for that in the near future I'll make a release with this.

dgonyeo:
Oh also saying it here in case anyone's curious: labels were added to the image-spec go library after v1.0.0-rc3, so acbuild won't be gaining support for manipulating those until another image-spec release.

vbatts:
@dgonyeo cool. We'll likely have an rc4 shortly

acbuild/acbuild.go
```diff
@@ -283,19 +317,22 @@ func stdout(format string, a ...interface{}) {
 }
 
 func addACBuildAnnotation(cmd *cobra.Command, args []string) error {
-	const annoNamePattern = "appc.io/acbuild/command-%d"
+	const annoNamePattern = "coreos.com/acbuild/command-%d"
```
s-urbaniak: I am a bit lost about this changed annotation name, why is this `coreos.com`?

runcom:
@cyphar would you also take a look at this since you've been working on umoci?

acbuild/layer.go
```diff
@@ -0,0 +1,57 @@
...
+
+	a, err := newACBuild()
+	if err != nil {
+		stderr("%v", err)
+		return 1
+	}
+	err = a.NewLayer()
+
+	if err != nil {
+		stderr("layer: %v", err)
```
s-urbaniak: nit: `creating layer failed: %v`

s-urbaniak: nevermind, I see you are consistent with this pattern below.

lib/appc/isolator.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return err
+	}
+	removeIsolatorFromMan(*acid, m.manifest)
+	m.manifest.App.Isolators = append(m.manifest.App.Isolators, *i)
+	return m.save()
+}
+
+// AddIsolator removes an isolator of name from the current manifest
+func (m *Manifest) RemoveIsolator(name string) error {
```
s-urbaniak: `RemoveIsolator removes ...`

lib/appc/manifest.go
```diff
@@ -0,0 +1,134 @@
...
+	var err error
+	if prettyPrint {
+		manblob, err = json.MarshalIndent(m.manifest, "", "    ")
+	} else {
+		manblob, err = m.manifest.MarshalJSON()
+	}
+	if err != nil {
+		return err
+	}
+	fmt.Println(string(manblob))
```
s-urbaniak: I guess this is a leftover ;-)

lib/begin.go
```diff
@@ -95,9 +119,13 @@ func (a *ACBuild) Begin(start string, insecure bool) (err error) {
 			case finfo.IsDir():
 				return a.beginFromLocalDirectory(start)
 			default:
-				return a.beginFromLocalImage(start)
+				return a.beginFromLocalImage(start, mode)
 			}
 		} else {
+			if mode == BuildModeOCI {
+				// TODO: fix this!
+				return fmt.Errorf("cannot start from remote OCI images currently")
```
s-urbaniak: once we merge this, we should create a follow-up tracking issue.

runcom: we'll should likely start using containers/image for this sort of images fetching. Eventually you can implement an appc:// transport in containers image and use something like "docker://busybox appc://busybox" to make the conversion directly.

lib/manifest.go
```diff
@@ -0,0 +1,381 @@
...
+}
+func (a *ACBuild) SetPreStart(exec []string) (err error) {
+	if err = a.lock(); err != nil {
+		return err
+	}
+	defer func() {
+		if err1 := a.unlock(); err == nil {
+			err = err1
+		}
+	}()
```
s-urbaniak: these lock/unlock logic is repeated a lot. can we dry this?

lib/oci/env.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	return i.save()
+}
+
+func (i *Image) removeFromEnv(name string) error {
+	foundOne := false
+	for j := len(i.config.Config.Env) - 1; j >= 0; j-- {
+		varParts := strings.Split(i.config.Config.Env[j], "=")
+		if len(varParts) != 2 {
+			return fmt.Errorf("invalid environment variable in config: %q", i.config.Config.Env[j])
```
s-urbaniak: is this something we forbid as per spec? my shell is perfectly happy with:
```
$ export foo=bar=baz
$ echo $foo
bar=baz
```

s-urbaniak: also if we err out here, `i.config.Config.Env` is in a broken state. for safety i'd create a copy and set it in case of success.

lib/oci/manifest.go
```diff
@@ -0,0 +1,283 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/containers/build/util"
+
+	//specs "github.com/opencontainers/image-spec/specs-go"
```
s-urbaniak: lefover?

lib/oci/manifest.go
```diff
@@ -0,0 +1,283 @@
...
+	pieces := strings.Split(hash, ":")
+	if len(pieces) != 2 {
+		return "", "", fmt.Errorf("error parsing digest in image ref: %q", hash)
+	}
+	return pieces[0], pieces[1], nil
+}
+
+func (i *Image) Close() error {
+	return nil
+}
```
s-urbaniak: can we document why we need to implement this method, I guess it's because of `io.Closer`?

lib/run.go
```diff
@@ -157,9 +166,101 @@ func (a *ACBuild) Run(cmd []string, workingDir string, insecure bool, runEngine
...
+	return deps, nil
+}
+
+func (a *ACBuild) generateOverlayPathsOCI() ([]string, error) {
+	var layerIDs []string
+	switch ociMan := a.man.(type) {
+	case *oci.Image:
+		layerIDs = ociMan.GetLayerHashes()
+	default:
+		return nil, fmt.Errorf("internal error: mismatched manifest type and build mode???")
```
s-urbaniak: let's remove the `???`

s-urbaniak: or is this even worth a `panic`?

lib/write.go
```diff
@@ -90,30 +99,46 @@ func (a *ACBuild) Write(output string, overwrite, sign bool, gpgflags []string)
...
+		defer aw.Close()
+		if err != nil {
+			pathErr, ok := err.(*os.PathError)
+			if !ok {
+				fmt.Printf("not a path error!\n")
+				return err
+			}
+			syscallErrno, ok := pathErr.Err.(syscall.Errno)
+			if !ok {
+				fmt.Printf("not a syscall errno!\n")
```
s-urbaniak: considering this is `lib` method, I'd rather suggest to inject any potential loggers.

s-urbaniak: or if this is a leftover ... then let's remov this and the above `fmt.Printf`

s-urbaniak:
@dgonyeo Huge kudos to this PR. I made a first pass on this and will also play around with it locally, looking great so far :+1: :+1:

lib/run.go
lucab: Please don't panic in library code.

lib/write.go
lucab: I made a similar comment in https://github.com/containers/build/pull/285#discussion_r96816258. I *think* error wrapping may be preferred here as well, as we just error-return after this.

acbuild/acbuild.go
dgonyeo: I changed it since acbuild doesn't only produce AppC images anymore. I can change it back if you think leaving `appc.io` annotations on OCI images would be fine.

lib/begin.go
dgonyeo: @runcom yup, I intend to use containers/image for fetching OCI images

lib/manifest.go
dgonyeo: I think I couldn't find a way to when I first wrote this since the deferred closure is using the fact that the named return variable is in scope to be able to check if the function returned with an error, and pulling that out into a named function would lose the closure properties, but I'll take another look.

dgonyeo: One option would be to silently drop errors from `a.unlock()`, but I dislike that for obvious reasons

lib/oci/env.go
dgonyeo: I think I meant to do `strings.SplitN(_, _, 2)`, so I'll go switch that over now.

lib/oci/manifest.go
dgonyeo: It's for the appc implementation of the `lib.Manifest` interface, which keeps an open handle to its manifest, but now that you've pointed this out I've noticed that this never gets called. I'm going to change things around so it's not needed, which will probably be easier than updating everywhere in the acbuild package to call Close.

lib/write.go
dgonyeo: Oh, these were debugging print statements I forgot to remove. My bad.

lib/appc/manifest.go
dgonyeo: How do you mean?

dgonyeo:
Addressed nits and suggestions, should be ready for another pass.

Oh, also a notable flaw in this PR I should mention is that currently all the tests for acbuild can only run against the appc build mode, since the way the tests work is by manually inspecting the manifest after asking acbuild to do something.

lib/appc/manifest.go
s-urbaniak: Sorry, it was a mechanical comment, in the context of this function it does make sense, you are right.

Nevertheless, since this lives in `lib`, I'd rather pass an `io.Writer` explicitely and use `fmt.Fprintln` here.

What do you think?

lib/run.go
s-urbaniak: ack, ignore my comment, let's not panic.

lib/appc/manifest.go
dgonyeo: Yup, makes sense. Just pushed the changes for this.

dgonyeo:
Oh, apparently the tests were failing after my last push. Just fixed that.

cyphar:
@runcom Sure. I was on holiday / at a conference so I've just been catching up on my backlog recently.

My first question for @dgonyeo is have you considered basing this off the library exposed inside https://github.com/cyphar/umoci -- to avoid duplicated effort and handle a bunch of edge cases.

dgonyeo:
@cyphar Hi! I chatted with you on IRC a couple months ago about combining our efforts, and what I got from the conversation was that you might be interested in the future, but not at that point in time.

Since then I've worked on adding this functionality into acbuild, which is what this is. It looks like if I were to incorporate umoci here it'd largely be helpful for manipulating manifests/configs in images, which is already finished in acbuild with this PR.

What extra edge cases do you think umoci will help handle?

dgonyeo:
bump on this

s-urbaniak:
LGTM

PR #285, opened 2017-01-17
lib/end: print a warning to run as root when permission denied

dgonyeo:
When a subcommand attempts to grab a lock due to permission denied or
end fails to delete something due to permission denied, a warning will
now be printed advising the user to run the command as root.

Example:
```
derek@proton ~/go/src/github.com/containers/build> ./bin/acbuild end
you'll need to run this as a user with appropriate privileges
end: remove .acbuild/currentaci/rootfs/var/www/localhost/htdocs: permission denied
```

Fixes https://github.com/containers/build/issues/284

### DIFF
--- a/lib/common.go
+++ b/lib/common.go
@@ -89,6 +89,15 @@ func (a *ACBuild) lock() error {
 
 	a.lockFile, err = os.OpenFile(a.LockPath, os.O_CREATE|os.O_RDWR, 0644)
 	if err != nil {
+		switch err1 := err.(type) {
+		case *os.PathError:
+			switch err2 := err1.Err.(type) {
+			case syscall.Errno:
+				if err2 == syscall.EACCES {
+					err = fmt.Errorf("permission denied: please run this as a user with appropriate privileges\n")
+				}
+			}
+		}
 		return err
 	}
 
--- a/lib/end.go
+++ b/lib/end.go
@@ -15,7 +15,9 @@
 package lib
 
 import (
+	"fmt"
 	"os"
+	"syscall"
 
 	"github.com/containers/build/util"
 )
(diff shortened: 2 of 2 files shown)


sanmai-NL:
One nitpick, please make it say ‘a user with appropriate privileges’ not ‘root’.

dgonyeo:
nitpick applied

lib/common.go
```diff
@@ -89,6 +89,15 @@ func (a *ACBuild) lock() error {
 
 	a.lockFile, err = os.OpenFile(a.LockPath, os.O_CREATE|os.O_RDWR, 0644)
 	if err != nil {
+		switch err1 := err.(type) {
+		case *os.PathError:
+			switch err2 := err1.Err.(type) {
+			case syscall.Errno:
+				if err2 == syscall.EACCES {
+					fmt.Fprintf(os.Stderr, "you'll need to run this as a user with appropriate privileges\n")
```
lucab: We already have errwrap in this lib, so I think it is better to wrap the original err. Libraries shouldn't directly printf to stderr.

lib/end.go
```diff
@@ -42,6 +44,15 @@ func (a *ACBuild) End() error {
 
 	err = os.RemoveAll(a.ContextPath)
 	if err != nil {
+		switch err1 := err.(type) {
+		case *os.PathError:
+			switch err2 := err1.Err.(type) {
+			case syscall.Errno:
+				if err2 == syscall.EACCES {
+					fmt.Fprintf(os.Stderr, "you'll need to run this as a user with appropriate privileges\n")
```
lucab: Same as above. Also, please turn the message into an impersonal form removing the "you", see https://github.com/coreos/docs/blob/master/STYLE.md#eschew-you

lib/common.go
dgonyeo: errwrap isn't currently used in acbuild, maybe it should be. The changes to incorporate it would be outside the scope of this PR I feel.

How's that look? Instead of printing to stderr I just generate a different error to return.

lucab: Ah, I saw it in glide.yaml thus I assumed we used it everywhere else. Manually wrapping/changing the error is also fine, it just needs a bit of care in not discarding the underlying error content.

dgonyeo:
@lucab this should be good to go if you want to give it another pass

PR #286, opened 2017-01-17
write: remove the --sign flag

dgonyeo:
Fixes https://github.com/containers/build/issues/252

For some definition of "fix"

### DIFF
--- a/acbuild/write.go
+++ b/acbuild/write.go
@@ -34,11 +34,16 @@ func init() {
 	cmdAcbuild.AddCommand(cmdWrite)
 
 	cmdWrite.Flags().BoolVar(&overwrite, "overwrite", false, "overwrite the resulting ACI")
-	cmdWrite.Flags().BoolVar(&sign, "sign", false, "sign the resulting ACI")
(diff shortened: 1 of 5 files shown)


acbuild/write.go
```diff
@@ -34,11 +33,10 @@ func init() {
 	cmdAcbuild.AddCommand(cmdWrite)
 
 	cmdWrite.Flags().BoolVar(&overwrite, "overwrite", false, "overwrite the resulting ACI")
-	cmdWrite.Flags().BoolVar(&sign, "sign", false, "sign the resulting ACI")
```
lucab: Should we perhaps be a bit more gentle here, still accepting the old flag and emitting a custom error for this specific case, instructing the user about direct gpg invocation?

dgonyeo: Done, it's a little gentler now:

```
derek@proton ~/go/src/github.com/containers/build> ./bin/acbuild write d --sign
write: the sign flag has been removed, please invoke gpg directly
```

dgonyeo:
Just updated this to remove mention of `--sign` from the docs.

PR #287, opened 2017-01-17
lib/run: better error output when dependency doesn't exist

dgonyeo:
If fetching a dependency for a run command, and the registry returns a
404 error, intercept that and print a more user-friendly error saying
that the dependency doesn't exist.

Example:
```
derek@proton 1 ~/go/src/github.com/containers/build> sudo ./bin/acbuild run -- apk update
run: dependency "quay.io/coreos/alpine-sh:doesnt-exist" doesn't appear to exist: bad HTTP status code: 404
```

Fixes https://github.com/containers/build/issues/283

### DIFF
--- a/lib/run.go
+++ b/lib/run.go
@@ -199,7 +199,12 @@ func (a *ACBuild) renderACI(insecure, debug bool) ([]string, error) {
 	var deplist []string
 	for _, dep := range man.Dependencies {
 		err := reg.FetchAndRender(dep.ImageName, dep.Labels, dep.Size)
-		if err != nil {
+		switch err {
+		case nil:
+		case registry.ErrNotFound:
+			l, _ := dep.Labels.Get("version")
+			return nil, fmt.Errorf("dependency %q doesn't appear to exist: %v", string(dep.ImageName)+":"+l, err)
+		default:
 			return nil, err
 		}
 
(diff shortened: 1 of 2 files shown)


dgonyeo:
I don't particularly like this implementation due to the `.Error() == "string"` check, but as far as I can tell the error at that point is just created via an [errors.New()](https://golang.org/pkg/errors/#example_New), so I don't know what else I have to work with.

I say this because `fmt.Printf("#%T\n")` resulted in an `*errors.errorString`.

lucab:
@s-urbaniak can you please have a review pass on this?

lib/run.go
```diff
@@ -200,6 +200,10 @@ func (a *ACBuild) renderACI(insecure, debug bool) ([]string, error) {
 	for _, dep := range man.Dependencies {
 		err := reg.FetchAndRender(dep.ImageName, dep.Labels, dep.Size)
 		if err != nil {
+			if err.Error() == "bad HTTP status code: 404" {
```
s-urbaniak: Hmm .. yes, this is not nice indeed. How about the following change which I quickly coded: https://github.com/s-urbaniak/build/commit/c9095272f3d91b9fddfbfcf8ad0a123d205e8a97?
It seems we already have an `ErrNotFound` error value in `lib`, so we can leverage that.

dgonyeo:
Yup, I like your implementation a lot better. Just pushed your changes.

PR #289, opened 2017-01-18
script: error if end is detected in a script

dgonyeo:
Fixes https://github.com/containers/build/issues/288


dgonyeo:
Decided to just make it error out instead of warn, it happens before any work is done so there won't be much time wasted on a user's part when they hit this.

PR #293, opened 2017-03-09
Handle host /etc/resolv.conf being a symlink

tjdett:
`/etc/resolv.conf` can be a symlink (eg. Fedora 25), but the symlink likely won't resolve inside the chroot. Copy `/etc/resolv.conf` as regular file instead.

Previous behaviour:
```
$ sudo "PATH=$(pwd)/bin:$PATH" bin/acbuild begin --debug ./library-alpine-3.5.aci
Beginning build with ./library-alpine-3.5.aci
$ sudo "PATH=$(pwd)/bin:$PATH" bin/acbuild run --engine chroot -- \
  sh -c 'ls -la /etc/resolv.conf && cat /etc/resolv.conf'
lrwxrwxrwx    1 root     root            35 Mar  7 22:22 /etc/resolv.conf -> /var/run/NetworkManager/resolv.conf
cat: can't open '/etc/resolv.conf': No such file or directory
```

New behaviour:
```
$ sudo "PATH=$(pwd)/bin:$PATH" bin/acbuild run --engine chroot -- \
  sh -c 'ls -la /etc/resolv.conf && cat /etc/resolv.conf'
-rw-r--r--    1 root     root           123 Mar  9 01:01 /etc/resolv.conf
# Generated by NetworkManager
...
```

### DIFF
--- a/engine/chroot/chroot.go
+++ b/engine/chroot/chroot.go
@@ -22,7 +22,6 @@ import (
 	"github.com/containers/build/engine"
 	"github.com/coreos/rkt/pkg/fileutil"
 	"github.com/coreos/rkt/pkg/multicall"
-	"github.com/coreos/rkt/pkg/user"
 )
 
 type Engine struct{}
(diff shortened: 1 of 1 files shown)


dgonyeo:
Awesome, thanks for the PR!

PR #297, opened 2017-03-27
oci: fix copying and symlinking

lucab:
This PR fixes two issues in OCI-mode `copy-to-dir` which were resulting in:
 * copies to target paths missing directory prefix
 * symlinks with empty target paths


lucab:
Tested locally with a simple package. You may want to run a quick smoke test just to double-check I didn't overlook anything.

PR #298, opened 2017-03-28
oci/manifest: use actual ImageLayer media type

lucab:
This PR fixes a bug where the mediaType for layers was incorrectly set to "MediaTypeImageLayer" literal.


dgonyeo:
Wow that's embarrassing.

PR #300, opened 2017-03-28
oci: fix blob-layer digest computation

lucab:
This fixes blob-layer digests in OCI images, by performing digest
computation on the final gzip'd layers content as specified by
current image-spec.


dgonyeo:
Cool, thanks!

vbatts:
:+1:

PR #301, opened 2017-03-29
oci: add timestamp to image config

lucab:


lucab:
For reference, this is an optional field which later turned into a `time.Time` field. This is giving me some issue during interop roundtrips due to https://github.com/opencontainers/image-spec/issues/626. Setting it explicitly here at source should workaround it without any side-effects.

PR #313, opened 2017-06-18
lib: show correct error when loadManifest fails

dalegaard:
The err value was accidentally getting overriden with the error from the
os.Stat call.


dgonyeo:
Thanks!

PR #314, opened 2017-06-18
Update OCI build mode, fix local dir start, and add remote capabilities

dalegaard:
These commits:

- Update the OCI build mode to output 1.0.0-rc6 compatible OCI images. The image-spec repo is updated as part of this.
- Fix the location of /etc/localtime when building under OCI mode. Before it simply ended up in rootfs, which is wrong for OCI mode.
- Make local directory starts work. This also used to save to the rootfs folder, so local directory has been turned into creating an empty start and then doing a copy-to-dir.
- Adds a remote start where acbuild will fetch the image over HTTP(S). Also prepared for a docker start, but haven't added it yet.

Since each part is pretty much reliant on the previous, I couldn't really split this into multiple pull requests. Let me know if I need to do anything different.

Thanks!


dalegaard:
I've added commits for some lazy rehashing code I wrote on top of these changes. This gives quite a considerable speedup on many workloads, so I'd rather add it now than risk forgetting about it. Sorry for piling on in a single PR.

PR #317, opened 2017-07-12
Error on trailing path separator in copy command

### LINKED ISSUE #201: acbuild copy does not consider targets ending in / as directories
It seems that "acbuild copy" does not consider a target path as a directory when it ends with / and thus the following two exhibit different behavior:

```
acbuild copy file /path/to/
acbuild copy file /path/to/file
```

The first one acts just like `acbuild copy file /path/to`

It seems that the problem happens because of this in lib/copy.go:CopyToTarget():

```
target := path.Join(a.CurrentACIPath, aci.RootfsDir, to)
```

since Join() removes any trailing / which then causes the subsequent path.Split() to return the last directory as a file.

sanmai-NL:
Resolves #201. See https://github.com/containers/build/issues/201#issuecomment-215243446.

### DIFF
--- a/acbuild/copy.go
+++ b/acbuild/copy.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"github.com/spf13/cobra"
+	"os"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


acbuild/copy.go
```diff
@@ -45,6 +46,13 @@ func runCopy(cmd *cobra.Command, args []string) (exit int) {
 		stderr("Copying host:%s to aci:%s", args[0], args[1])
 	}
 
+	if args[1][len(args[1])-1] == os.PathSeparator {
+		stderr(`There is a trailing path separator in the second 
+			operand. Please review the differences between 
+			this command and copy-to-dir. `)
+		return 1
+	}
+
```
dgonyeo: I think this will panic if `args[1]` is `""`, can you fix that?

sanmai-NL: Solved.

dgonyeo:
If you can squash the commits I'll merge

sanmai-NL:
@dgonyeo: done.

PR #318, opened 2017-07-12
Clarify interaction `resolv.conf` and engine

### LINKED ISSUE #153: acbuild should remove /etc/resolv.conf
I'm trying to use ubuntu-14.04-server-cloudimg-amd64-root as the base for my ACI, but I am unable to resolve host names when using `acbuild run`. It looks like systemd-nspawn is trying to populate `/etc/resolv.conf`, but is getting confused because that file is a symlink. I think it should be safe for acbuild to ensure that that file doesn't exist by the time that systemd-nspawn is run.

sanmai-NL:
May suffice to resolve #153.

### DIFF
--- a/Documentation/subcommands/run.md
+++ b/Documentation/subcommands/run.md
@@ -63,8 +63,6 @@ involved, so the command will be able to see and possibly interact with other
 processes on the host. This engine notably has no dependency on systemd, unlike
 the `systemd-nspawn` engine.
 
-
-
 ### Exiting out of systemd-nspawn
 
 All acbuild commands can be cancelled with Ctrl+c with the exception of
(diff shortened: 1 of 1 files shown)


Documentation/subcommands/run.md
```diff
@@ -39,21 +39,32 @@ squash` is being worked on to do this.
...
+### `chroot`
 
 An alternative engine is called `chroot`, which uses the chroot syscall to
 enter into the container and run the specified command. There's no namespacing
 involved, so the command will be able to see and possibly interact with other
 processes on the host. This engine notably has no dependency on systemd, unlike
 the `systemd-nspawn` engine.
 
+
+
```
dgonyeo: Can you remove these two newlines?

sanmai-NL:
@dgonyeo: done

PR #319, opened 2017-07-12
Warn when begin command isn’t run as superuser

### LINKED ISSUE #272: acbuild begin silently does nothing when run not by root
I created rootfs folder with debootstrap, played with acbuild with no luck. Spent a lot of time before realised that "acbuild begin" actually copies files under .acbuild/currentaci/ and it fails to do so unless run under root user.
I assume this is because it tries to keep file ownership and just cannot create root-owned files. But it definitely makes sense to alert user when you cannot do something not just swallow problems. To me, failure to copy a single file already warrants non-zero exit code. Maybe with a command line option to change that behaviour.
Cheers

sanmai-NL:
Resolves #272.

### DIFF
--- a/lib/begin.go
+++ b/lib/begin.go
@@ -73,6 +73,12 @@ func (a *ACBuild) Begin(start string, insecure bool, mode BuildMode) (err error)
 	if err = a.lock(); err != nil {
 		return err
 	}
+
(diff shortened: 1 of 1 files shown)


sanmai-NL:
@dgonyeo: ready to merge?

PR #320, opened 2017-07-12
Improve shell script

sanmai-NL:
* Use a POSIX-compatible shell script and remove references to Bash.
* More complete handling of corner cases (e.g. failing `acbuild end`, termination signal, etc.)
* Mark tutorial-style info as extra background section.

### DIFF
--- a/Documentation/shell-scripts.md
+++ b/Documentation/shell-scripts.md
@@ -1,39 +1,40 @@
-# Bash scripts
+# Managing `acbuild` using shell scripting
 
-Bash scripting provides a common, simple, and well-known mechanism to drive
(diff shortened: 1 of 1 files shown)


dgonyeo:
What's the motivation against using bash? It should be widely available, and has more features for users when writing scripts.

sanmai-NL:
Bash is not available by default under e.g. Alpine Linux. The sample code does not require Bash in any way. [Bash has had CVEs](https://www.cvedetails.com/product/21050/GNU-Bash.html?vendor_id=72), more than e.g. [dash](https://www.cvedetails.com/vulnerability-list/vendor_id-9585/Dash.html). Bash may be nice for interactive shells, for shell scripting a portable script is better.

